PR #4, opened 2016-04-27
network

diasdavid:
This is a **WIP**, missing: 

- [x] Update Swarm with `peer-mux-*` events
- [x] Update peerBook on js-ipfs to make it smarter when it comes to updating the info about a peer (with the option to flush it to disk) and indexing. So that I can fetch the peerInfo for `sendMessage` call
- [x] Update Swarm to patch the peerId to the conn (and the streams that are multiplexed on the conn), so that we can do the `_receiveMessage`

### DIFF
--- a/src/network/index.js
+++ b/src/network/index.js
@@ -1,18 +1,68 @@
 'use strict'
 
+const bl = require('bl')
+const fs = require('fs')
+const path = require('path')
+const protobuf = require('protocol-buffers')
+const pbm = protobuf(fs.readFileSync(path.join(__dirname, '../message/message.proto')))
+
 module.exports = class Network {
-  constructor (libp2p) {
-    // TODO: Implement me
+  constructor (libp2p, peerBook, bitswap) {
     this.libp2p = libp2p
+    this.peerBook = peerBook
+
+    this.libp2p.swarm.handle('/ipfs/bitswap/1.0.0', (conn) => {
+      conn.pipe(bl((err, data) => {
+        conn.end()
+        if (err) {
+          return bitswap._receiveError(err)
+        }
+        let msg
+        try {
+          msg = pbm.Message.decode(data)
+        } catch (err) {
+          return bitswap._receiveError(err)
+        }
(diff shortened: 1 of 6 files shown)


src/network/index.js
```diff
@@ -1,14 +1,38 @@
...
   }
 
   // Connect to the given peer
   connectTo (peerId, cb) {
-    // TODO: Implement me
+    // NOTE: For now, all this does is ensure that we are
+    // connected. Once we have Peer Routing, we will be able
+    // to find the Peer
+    if (this.libp2p.swarm.muxedConns[peerId.toB58String()]) {
+      // TODO: ask - want _onPeerConnected to be called here too?
```
diasdavid: @dignifiedquire should I call _onPeerConnected here too? Or is the `cb` enough to signal we successfully  dialed (I'm afraid that we fire to routines for connecting to the same peer)

dignifiedquire:
@diasdavid can you rebase please?

src/network/index.js
dignifiedquire: I need to check

diasdavid:
@dignifiedquire rebased, but do not merge yet :)

src/network/index.js
dignifiedquire: No need to call `_onPeerConnected` here, as that will already have been called from the first connection in this case.

diasdavid:
@dignifiedquire the network is implemented :)

test/index-test.js
```diff
@@ -146,7 +150,10 @@ module.exports = (repo) => {
       })
     })
 
-    it('block is retrived from peer', (done) => {
+    // Not sure if I understand what is going on here
+    // test fails because now the network is not properly mocked
+    // what are these net.stores and mockNet.bitswaps?
```
diasdavid: @dignifiedquire could you clarify what is going on in this test? libp2p needs to be properly mocked, but I couldn't get what was exactly was going on/intended.

dignifiedquire: Yes, I need to change the mock network tool to now mock libp2p instead of the network instance

PR #7, opened 2016-05-05
fix(network): better message handling

dignifiedquire:
Some cleanup and correct usage of the `Message`

### DIFF
--- a/src/network/index.js
+++ b/src/network/index.js
@@ -1,68 +1,82 @@
 'use strict'
 
 const bl = require('bl')
-const fs = require('fs')
-const path = require('path')
-const protobuf = require('protocol-buffers')
-const pbm = protobuf(fs.readFileSync(path.join(__dirname, '../message/message.proto')))
+const async = require('async')
+const Message = require('../message')
 
 module.exports = class Network {
   constructor (libp2p, peerBook, bitswap) {
     this.libp2p = libp2p
     this.peerBook = peerBook
+    this.bitswap = bitswap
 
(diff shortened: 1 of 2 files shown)


dignifiedquire:
@diasdavid ^^

src/network/index.js
```diff
@@ -1,68 +1,82 @@
...
-    this.libp2p.swarm.on('peer-mux-established', (peerInfo) => {
-      bitswap._onPeerConnected(peerInfo.id)
-    })
+  _attachSwarmListeners () {
+    this.libp2p.swarm.handle('/ipfs/bitswap/1.0.0', this._onConnection.bind(this))
 
-    this.libp2p.swarm.on('peer-mux-closed', (peerInfo) => {
-      bitswap._onPeerDisconnected(peerInfo.id)
-    })
+    this.libp2p.swarm.on('peer-mux-established', this._onPeerMux.bind(this))
```
diasdavid: This changes what gets passed to the function to become a peerInfo instead of a peerId (as you had specified on the interface/stub)

src/network/index.js
```diff
@@ -1,68 +1,82 @@
...
+
+    this.libp2p.swarm.on('peer-mux-closed', this._onPeerMuxClosed.bind(this))
```
diasdavid: Same here

src/network/index.js
diasdavid: Ah, saw it wrong, you defined a new method. Ok

diasdavid:
LGTM :)

PR #10, opened 2016-05-05
feat: add unwantBlocks method

dignifiedquire:


src/index.js
```diff
@@ -137,9 +137,9 @@ module.exports = class Bitwap {
       cb(err, block)
     }
 
-    this.getBlocks([key], (err, res) => {
-      if (err) {
-        return done(err)
+    this.getBlocks([key], (errs, res) => {
+      if (errs) {
+        return done(errs[0])
```
diasdavid: What about all the other errs?

diasdavid: Maybe do like the `block-service`? Where getBlocks returns an object of `mutlihash: { err, block} `

src/index.js
```diff
@@ -137,9 +137,9 @@ module.exports = class Bitwap {
...
       }
 
       done(null, res[0])
```
diasdavid: Same here, we just want one block?

src/index.js
dignifiedquire: There should be max on error here, but could do the object approach

src/index.js
dignifiedquire: the method is called `getBlock` so yes we only want a single block, and will only receive back a single block.

src/index.js
diasdavid: +1 for object approach

dignifiedquire: @diasdavid done, please give this another round of review so I can merge it soon :)

API.md
```diff
@@ -25,7 +25,22 @@ Fetch a single block.
 - `keys: []Multihash`
 - `cb: Function`
 
-Fetch multiple blocks.
+Fetch multiple blocks. The `cb` is called with a result object of the form
+```js
+{
+  [key1]: {error: errorOrUndefined, block: blockOrUndefined},
```
diasdavid: "key being the multihash of the block"

src/index.js
```diff
@@ -138,12 +138,11 @@ module.exports = class Bitwap {
...
-    this.getBlocks([key], (err, res) => {
-      if (err) {
-        return done(err)
-      }
+    this.getBlocks([key], (results) => {
+      const err = results[key].error
+      const block = results[key].block
 
-      done(null, res[0])
+      done(err, block)
```
diasdavid: Not sure if I understand what is going on here, is this some es6 witchcraft?

src/index.js
```diff
@@ -153,15 +152,48 @@ module.exports = class Bitwap {
...
+        cleanupListeners()
+        cb(results)
       }
     }
 
+    addListeners()
+
     keys.forEach((key) => {
       // Sanity check, we don't want to announce looking for blocks
       // when we might have them ourselves
```
diasdavid: This is more than sanity, since now the Block Service has an 'online' mode that defers everything to bitswap

test/index-test.js
```diff
@@ -120,170 +120,203 @@ module.exports = (repo) => {
...
-          expect(res).to.be.eql(res)
-          cb()
+      it('block is sent after local add', (done) => {
+        const me = PeerId.create({bits: 64})
+        const other = PeerId.create({bits: 64})
+        const block = new Block('hello world local add')
+        let bs1
+        let bs2
+        let n1
+        let n2
```
diasdavid: You don't reuse this vars, they can be `const`

src/index.js
dignifiedquire: No witchcraft, what do you not understand? If there is no `err` it 'll be passed as `done(undefined, block)` and otherwise the `err` will be the first arg

diasdavid: ah, nvm, I misread and thought this was a function declaration, I was wondering that a array argument was and why only return one block in a func that is called getBlocks 

Thank you :)

PR #11, opened 2016-05-06
WIP - gen bitswap network

diasdavid:
- [x] func to generate random bitswap networks
- [x] rebase latest @dignifiedquire changes
- [x] confirm that swarm is ok with dialing no protocol at all


test/network/gen-bitswap-network.node.js
```diff
@@ -0,0 +1,67 @@
...
+      // setTimeout is used to avoid closing the TCP socket while spdy is
+      // still sending a ton of signalling data
+      const tasks = nodeArr.map((node) => {
+        return (cb) => {
+          node.libp2p.swarm.close(cb)
+        }
+      })
+
+      setTimeout(() => {
+        parallel(tasks, done)
```
dignifiedquire: Lets use `async.parallel` instead of yet another module in here

dignifiedquire: I've been using `run-parallel` only in places where `async` was not used before.

test/network/gen-bitswap-network.node.js
```diff
@@ -0,0 +1,67 @@
...
+      }, 1000)
```
dignifiedquire: Would be good to have a comment as to why this timeout exists

dignifiedquire: nvm, saw the comment now

test/network/gen-bitswap-network.node.js
diasdavid: And I thought I would make you happy to stick to your new trend of `run-parallel` :D

test/utils.js
```diff
@@ -84,3 +91,87 @@ exports.applyNetwork = (bs, n) => {
...
+  function createBitswaps () {
+    netArray.forEach((net) => {
+      net.bitswap = new Bitswap(net.peerInfo, net.libp2p, net.repo, net.peerBook)
+    })
+    establishLinks()
+  }
+
+  // connect all the nodes between each other
+  function establishLinks () {
+    async.eachLimit(netArray, 1, (from, cbI) => {
```
dignifiedquire: you can just use `async.series` if you want to limit to `1`

test/utils.js
```diff
@@ -3,7 +3,14 @@
...
 const _ = require('lodash')
 const PeerId = require('peer-id')
+const PeerInfo = require('peer-info')
+const PeerBook = require('peer-book')
+const multiaddr = require('multiaddr')
 const Bitswap = require('../src')
+const libp2p = require('libp2p-ipfs')
+const os = require('os')
+const Repo = require('ipfs-repo')
+const fsBS = require('fs-blob-store')
```
dignifiedquire: Maybe it's more efficient if we use the memory blob store for these tests, what do you think?

test/utils.js
diasdavid: good point, changing

diasdavid:
@dignifiedquire applied your CR

test/network/gen-bitswap-network.node.js
dignifiedquire: sorry :D

PR #12, opened 2016-05-06
fix: issues found when integrating into js-ipfs

dignifiedquire:


src/decision/engine.js
```diff
@@ -78,11 +79,11 @@ module.exports = class Engine {
   }
 
   wantlistForPeer (peerId) {
-    if (!this.ledgerMap.has(peerId)) {
+    if (!this.ledgerMap.has(peerId.toB58String())) {
```
diasdavid: 👍

diasdavid:
LGTM :D

Issue #8, opened 2016-05-05
Create bitswap Network for Testing


diasdavid:
Fire N number of nodes and connect all of them together

dignifiedquire:
Pasting the notes from the etherpad. Later we should do also the same tests on full js-ipfs nodes, so that we account for every possibility

**bitswap tests**
- [ ] script to spawn a node, given a config file, creates a random file every X seconds and prints the hash (so that it can be requested
- spawn 1 node on the main thread with TCP
- dial to the other down
- fetch all of the blocks the other node outputs
- [ ] spawn 5 nodes in separate process and repeat the above
- [ ] spawn 10 nodes and ..
- [ ] spawn 20 nodes
- repeat the previous proc
- note: measure time (avg, max and min) of fetching the block and memory usage in the beginning and end

timing data in different events
  - use logs (debug)

dignifiedquire:
In #17 there was a set of tests added that are skipped currently as we are seeing issues on CI with these. Need to figure out what's wrong and fix them.

dignifiedquire:
I have enabled them in #25, lets see if we can get them to pass.

dignifiedquire:
They run on CI now, only for 2 nodes, but it is something 😄

PR #17, opened 2016-05-29
Stronger faster better

dignifiedquire:
Found issues when distributing blocks in networks with `>= 3` nodes. This adds tests for that and hopefully a fix soon :)

cc @diasdavid @haadcode

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,23 +40,26 @@
     "fs-blob-store": "^5.2.1",
     "idb-plus-blob-store": "^1.1.2",
     "ipfs-repo": "^0.8.0",
-    "libp2p-ipfs": "^0.10.0",
-    "lodash": "^4.11.2",
+    "libp2p-ipfs": "^0.11.0",
+    "lodash": "^4.13.1",
     "multiaddr": "^2.0.2",
     "ncp": "^2.0.0",
     "peer-book": "^0.3.0",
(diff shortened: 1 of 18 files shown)


dignifiedquire:
Mostly works, but `socket hung up` errors are making life hard :/

diasdavid:
Pasting the notes from the etherpad. Later we should do also the same tests on full js-ipfs nodes, so that we account for every possibility

**bitswap tests**
- [ ] script to spawn a node, given a config file, creates a random file every X seconds and prints the hash (so that it can be requested
- spawn 1 node on the main thread with TCP
- dial to the other down
- fetch all of the blocks the other node outputs
- [ ] spawn 5 nodes in separate process and repeat the above
- [ ] spawn 10 nodes and ..
- [ ] spawn 20 nodes
- repeat the previous proc
- note: measure time (avg, max and min) of fetching the block and memory usage in the beginning and end

timing data in different events
  - use logs (debug)

dignifiedquire:
@diasdavid as not all of this will happen in this PR I moved the notes to #8

diasdavid:
Comment this line https://github.com/ipfs/js-ipfs-bitswap/pull/17/files#diff-ff2899c03c9e8e31a9b69d7621e08c83R109 :)

PR #19, opened 2016-06-07
fix: use length-prefixed-stream for messages for go interop

dignifiedquire:
cc @diasdavid @whyrusleeping

### DIFF
--- a/src/message/index.js
+++ b/src/message/index.js
@@ -45,19 +45,25 @@ class BitswapMessage {
   }
 
   toProto () {
-    return pbm.Message.encode({
+    const msg = {
       wantlist: {
         entries: Array.from(this.wantlist.values()).map((e) => {
           return {
-            block: mh.toB58String(e.key),
+            block: e.key,
             priority: Number(e.priority),
             cancel: Boolean(e.cancel)
           }
-        }),
-        full: this.full
+        })
       },
-      blocks: Array.from(this.blocks.values()).map((b) => b.data)
-    })
+      blocks: Array.from(this.blocks.values())
+        .map((b) => b.data)
+    }
+
+    if (this.full) {
+      msg.wantlist.full = true
+    }
+
+    return pbm.Message.encode(msg)
   }
 
   equals (other) {
@@ -89,7 +95,7 @@ BitswapMessage.fromProto = (raw) => {
   const m = new BitswapMessage(dec.wantlist.full)
 
   dec.wantlist.entries.forEach((e) => {
-    m.addEntry(mh.fromB58String(e.block), e.priority, e.cancel)
+    m.addEntry(e.block, e.priority, e.cancel)
   })
   dec.blocks.forEach((b) => m.addBlock(new Block(b)))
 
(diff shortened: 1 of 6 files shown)


diasdavid:
@dignifiedquire the other thing is that block key on the protobuf is no B58, it is the multihash bytes

src/message/index.js
```diff
@@ -49,7 +49,7 @@ class BitswapMessage {
       wantlist: {
         entries: Array.from(this.wantlist.values()).map((e) => {
           return {
-            block: mh.toB58String(e.key),
+            block: e.key.toString(),
```
diasdavid: `s/e.key.toString()/e.key.toString('binary')`

src/message/index.js
```diff
@@ -89,7 +89,7 @@ BitswapMessage.fromProto = (raw) => {
   const m = new BitswapMessage(dec.wantlist.full)
 
   dec.wantlist.entries.forEach((e) => {
-    m.addEntry(mh.fromB58String(e.block), e.priority, e.cancel)
+    m.addEntry(new Buffer(e.block, 'utf8'), e.priority, e.cancel)
```
diasdavid: s/utf8/binary

test/message.spec.js
```diff
@@ -43,7 +56,7 @@ describe('BitswapMessage', () => {
     const raw = pbm.Message.encode({
       wantlist: {
         entries: [{
-          block: mh.toB58String(new Buffer('hello')),
+          block: new Buffer('hello'),
```
diasdavid: `.toString('binary')`

avoid using a valid utf8 string here because otherwise the protobuf won't catch it. Put some random bytes

dignifiedquire:
Looks like `protocol-buffers` does something different than the official go implementation for strings in some cases. Not sure what is going wrong though :(

dignifiedquire:
So I have figured out why we are not able to properly encode things into the protobuf and I'm afraid it is a non spec compliant use of protobufs from go.
According to the spec for protobufs a `string` field must be a sequence of bytes that are valid `utf8` sequences. It turns out this is not the case for multihashes as raw bytes, which is what go puts into this field.

The reason why we are hitting issues is because in JavaScript our library actually ensures this property resulting in a replacement of all non valid utf8 bytes to the sequence `U+FFFD`. You can see this when converting one of these from buffer to string and back in JavaScript:

```js
const mh = require('multihashes')
const hash = mh.fromB58String('QmRN6wdp1S2A5EtjW9A3M1vKSBuQQGcgvuhoMUoEz4iiT5')

console.log(hash.toString('hex'))
// => 12202cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824

const converted = new Buffer(hash.toString('utf8'), 'utf8')
console.log(converted.toString('hex'))
// => 12202cefbfbd4defbfbd5fefbfbdefbfbd0e26efbfbd3b2ac5b9efbfbdefbfbd1b161e5c1fefbfbd425e73043362efbfbdefbfbdefbfbd24
```

dignifiedquire:
getting closer I know have this

```
js: CiwKKAoiEiAs8k26X7CjDiboOyrFueKeGxYeXB+nQl5zBDNik4uYJBAKGAAQAA==
go: CioKKAoiEiAs8k26X7CjDiboOyrFueKeGxYeXB+nQl5zBDNik4uYJBAKGAA=
```

diasdavid:
@dignifiedquire I see you merged this one. Is it fully fixed?

dignifiedquire:
@diasdavid yes :)

PR #25, opened 2016-08-21
[WIP] Pull streams

dignifiedquire:
- [x] New pull-stream based interface used in `js-ipfs-block-service`
- [x] Tests passing with all deps moved to pull-streams
- [x] Modular async
- [x] replace highland.js with pull-stream


.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
 language: node_js
 node_js:
   - 4
-  - 5
+  - stable
```
diasdavid: Let's add Node.js 5 too

API.md
```diff
@@ -6,35 +6,31 @@
...
-
-Fetch a single block.
+### `getStream(key)`
 
-> Note: This is safe guarded so that the network is not asked
-> for blocks that are in the local `datastore`.
+- `key: Multihash`
 
-### `getBlocks(keys, cb)`
+Returns a source `pull-stream`. Values emitted are the received blocks.
```
diasdavid: Is this going to return several blocks for one key? Seems that there is a need for get(key), get(keys), getStream(keys)

API.md
```diff
@@ -6,35 +6,31 @@
...
+  })
+)
 ```
 
-Where `key<i>` is the multihash of the block.
+> Note: This is safe guarded so that the network is not asked
+> for blocks that are in the local `datastore`.
+
 
 ### `unwantBlocks(keys)`
```
diasdavid: since get is not getblock, let's make this just `unwant`

API.md
```diff
@@ -50,6 +46,9 @@ attached to `getBlock` are errored with `Error('manual unwant: key')`.
 
 Cancel previously requested keys.
 
+### `putStream()`
+
+Returns a duplex `pull-stream` that emits an object `{key: Multihash}` for every written block when it was stored.
 
 ### `hasBlock(block, cb)`
```
diasdavid: same goes for `s/hasBlock/has`. However, should bitswap have this call? What would be the case?

package.json
```diff
@@ -33,40 +33,43 @@
   },
   "homepage": "https://github.com/ipfs/js-ipfs-bitswap#readme",
   "devDependencies": {
-    "abstract-blob-store": "^3.2.0",
-    "aegir": "^3.2.0",
+    "aegir": "^6.0.1",
```
diasdavid: let's bump to aegir 8.0.0

diasdavid:
I would appreciate (and all of the contributors) if there was some documentation explaining how bitswap works internally, currently it requires to dig through the code to confirm every detail.

API.md
dignifiedquire: `hasBlock` is called when a new block arrives

API.md
dignifiedquire: fixed, it now takes an array, or a single key.

API.md
dignifiedquire: done

API.md
dignifiedquire: I don't like `has`, it is a bit cryptic in this case. Any better suggestions?

dignifiedquire: Maybe `add`?

package.json
dignifiedquire: done

API.md
diasdavid: `has` is called when the block arrives? does it add the block? Shouldn't it just be put then?

dignifiedquire: `put` makes sense to me, yes it adds the block to the storage + tells the other bitswap parts that it has this block now. I copied the naming from the go implementation, that's where this comes from.

diasdavid: I thought it was just a check. Let's just use `put` for this.

dignifiedquire: done

dignifiedquire:
Latest failure seems only due to browserify-zlib not being the right version

diasdavid:
@dignifiedquire https://github.com/ipfs/js-ipfs-bitswap/pull/25#issuecomment-245755678 why is this a problem only in js-ipfs-bitswap?

dignifiedquire:
Tests are actually passing on travis, for some reason though aegir doesn't agree with that -.-

PR #41, opened 2016-10-03
Async Crypto Endeavour

dignifiedquire:

PR #43, opened 2016-10-12
Tell browserify to fix up fs.readFileSync calls

interfect:
@substack suggested that this would make browserify able to actually package this module. Without running through brfs, it packages up just fine, but browserify provides an empty `fs` implementation, so when this module goes to load the .proto files on startup it fails.


interfect:
Looks like the network "before all" hook test is timing out. I suspect it's not my fault.

dignifiedquire:
Thank you but not needed anymore with #41

PR #45, opened 2016-10-20
Bitswap Complete (1.0.0 + 1.1.0)

diasdavid:
> Nothing to see here yet, just created to keep tabs and references in what needs to be done
- [ ] Needs to support both bitswap 1.1.0 and 1.0.0
- [x] Needs js-multistream to support custom validation functions (mainly semver)
- [ ] New Bitswap message.proto (supports both old and new)
  - https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/message/pb/message.proto

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,7 +16,7 @@ const WantManager = require('./wantmanager')
 const Network = require('./network')
 const decision = require('./decision')
 
-module.exports = class Bitwap {
+class Bitswap {
   constructor (p, libp2p, blockstore, peerBook) {
     // the ID of the peer to act on behalf of
     this.self = p
@@ -32,6 +32,7 @@ module.exports = class Bitwap {
     // handle message sending
     this.wm = new WantManager(this.network)
 
+    // stats
     this.blocksRecvd = 0
     this.dupBlocksRecvd = 0
     this.dupDataRecvd = 0
@@ -42,7 +43,7 @@ module.exports = class Bitwap {
 
   // handle messages received through the network
   _receiveMessage (peerId, incoming, cb) {
-    cb = cb || (() => {})
+    cb = cb || noop
     log('receiving message from %s', peerId.toB58String())
     this.engine.messageReceived(peerId, incoming, (err) => {
       if (err) {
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -304,3 +310,7 @@ module.exports = class Bitwap {
     this.engine.stop()
   }
 }
+
+module.exports = Bitswap
+
+function noop () {}
```
dignifiedquire: didn't we agree that we would stop doing style refactorings that are based on personal preferences?

src/message/index.js
```diff
@@ -35,8 +35,9 @@ class BitswapMessage {
     }
   }
 
-  addBlock (block) {
-    this.blocks.set(mh.toB58String(block.key), block)
+  addBlock (block, hashAlg) {
+    const key = block.key(hashAlg)
```
dignifiedquire: this is going to be async with the new crypto. so might be better to base it on top of that

test/browser.js
```diff
@@ -25,8 +25,8 @@ function createRepo (id, done) {
     })
   })
 
-  const mainBlob = new Store(id)
-  const blocksBlob = new Store(`${id}/blocks`)
+  const mainBlob = new Store(id + Math.random())
+  const blocksBlob = new Store(`${id}/blocks` + Math.random())
```
dignifiedquire: why?

test/index-test.js
```diff
@@ -46,7 +46,7 @@ module.exports = (repo) => {
     })
 
     describe('receive message', () => {
-      it('simple block message', (done) => {
+      it.only('simple block message', (done) => {
```
dignifiedquire: TODO: remove before merge

src/index.js
diasdavid: I might be wrong, but I don't see this as a style preference. Seeing `cb = cb || noop` is more simpler than `cb = cb || (() => {})` everytime I want a noop.

I can revert back though.

As I said at the top of the PR: "Nothing to see here", which was meant to save time from CR, this is far from complete (note that I didn't add the "needs review" label).

dignifiedquire:
> nothing to see here

...someone needs to pay better attention to headlines *cough Just ignore me till you are ready :)

test/browser.js
diasdavid: This was due to https://github.com/ipfs/js-ipfs/issues/528#issuecomment-255138513

PR #67, opened 2016-11-11
fix: better accounting and less logs

dignifiedquire:
The last fixes needed for https://github.com/ipfs/js-ipfs/pull/485

### DIFF
--- a/src/decision/engine.js
+++ b/src/decision/engine.js
@@ -66,7 +66,7 @@ module.exports = class Engine {
         return Boolean(nextTask)
       },
       (next) => {
-        log('got task', nextTask)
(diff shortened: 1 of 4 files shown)


dignifiedquire:
@diasdavid can I get some review on this?

src/decision/engine.js
```diff
@@ -66,7 +66,7 @@ module.exports = class Engine {
         return Boolean(nextTask)
       },
       (next) => {
-        log('got task', nextTask)
+        log('got task')
```
dignifiedquire: There is already so much logging, that's fine. and this version was printing the whole struct which was complete overkill

PR #76, opened 2016-12-10
WIP: Support bitswap 1.1.0 and bitswap 1.0.0 using CID

diasdavid:
#### tl;dr; This PR adds support to CID and enables js-ipfs-bitswap to interop with both bitswap versions.

My initial goal was only to add the new support, however, I was having  trouble to deal with the cognitive load that I was getting by having to sync in how ipfs-bitswap works and how this repo is organized each time I context switched. Since there was never a [documentation/writeup](https://github.com/ipfs/js-ipfs/issues/17#issuecomment-239774235), I started adding notes, later an image of the architecture and finally organized a bit the folder structure, so that the structure gives us a more intuitive way to understand what is going on. I hope this helps other people contribute and debug bitswap.

## Thoughts and questions to answer

Also, during the process, I've realized that some things can be improved, that some others look like 'Go code being forced in JS land' and some that I'm simply not sure why it is the way it is. It doesn't have to be part of this PR, but I want to make sure to write them down so that we go through them later. First, here is the:

image

- I don't see any code to coalesce messages in the messages queues, which means the more peers we have, the more buffering it will happen, to the point where the message to reset the wantlist might be added even before the previous having been sent and nothing is clearing the message queue.
- The decision engine buffers the blocks in memory when it informs the message queue that there is one more block to send, this means that if thousands of transfers are being performed, every block will stay in memory till the message gets sent. A better approach would be to only read from disk when the message is going to be sent.
- Why is there a refcounter for PeerHandler in the WantManager?
- Is a MessageQueue per peer the best approach? Why not piggy back on want list updates and just check what the peer wants and attach the blocks necessary?
- Why have two classes for WantListEntry? The BitSwapMessage Entry only has one more property `cancel`, can't we just `extend` the other and add that property?
- stats (exchanged byteCount and so on) kind of don't have any self guards to overflowing

## What's left

- [x] wantlist cid support
- [x] BitSwapMessage cid support
  - [x] .serializeToBitswap1.0.0
  - [x] .serializeToBitswap1.1.0
- [x] MessageQueue cid support
- [x] WantManager cid support
- [x] Create pretty graph
- [x] Update test folder structure to reflect moving around in src
- [x] DecisionEngine cid support
  - [x] PriorityQueue
  - [x] PeerRequestQueue
  - [x] ledger
  - [x] engine
- [x] Create network transform that serializes bitswap to 1.0.0 instead of 1.1.0
- [x] Default to dial on the new protocol version, try again on 1.0.0 if it fails


dignifiedquire:
this is super hard to review with all the file moving and renaming. Can you please put that into a different PR?

test/wantmanager/index.spec.js
```diff
@@ -12,25 +12,26 @@ const Wantmanager = require('../../src/wantmanager')
...
+      const peer2 = peerIds[1]
 
       let wm
+
       const network = mockNetwork(6, (calls) => {
         expect(calls.connects).to.have.length(6)
         const m1 = new Message(true)
         m1.addEntry(new Buffer('hello'), cs.kMaxPriority)
-        m1.addEntry(new Buffer('world'), cs.kMaxPriority - 1)
+        m1.addEntry(ew Buffer('world'), cs.kMaxPriority - 1)
```
dignifiedquire: I don't think this was a good change 😛

src/wantmanager/index.js
```diff
@@ -12,10 +11,10 @@ const MsgQueue = require('./msg-queue')
 const log = debug('bitswap:wantmanager')
 log.error = debug('bitswap:wantmanager:error')
 
-module.exports = class Wantmanager {
+module.exports = class WantManager {
```
dignifiedquire: :(

dignifiedquire:
@diasdavid can we please move the question and answer section into individual issues, I think those all warrant discussion/thought/individual PRs and should be handled individually.

dignifiedquire:
@diasdavid 😍 thank you for writing that diagram

dignifiedquire:
Can you please update the test folder structure so it reflects the one on src?

diasdavid:
> @diasdavid can we please move the question and answer section into individual issues, I think those all warrant discussion/thought/individual PRs and should be handled individually.

Absolutely, that is my plan once this gets merged. For now, it is just a place for me to collect everything.

> @diasdavid 😍 thank you for writing that diagram

Glad you like it :D 

> Can you please update the test folder structure so it reflects the one on src?

On it, added to the tasks

src/wantlist/index.js
```diff
@@ -51,8 +55,9 @@ class Wantlist {
     return new Map(Array.from(this.set.entries()).sort())
   }
 
-  contains (key) {
-    return this.set.get(mh.toB58String(key))
+  contains (cid) {
+    const cidStr = cid.toBaseEncodedString()
+    return !!this.set.get(cidStr)
```
diasdavid: Oh, you want it to continue returning the value? ok

test/wantmanager/index.spec.js
diasdavid: `ew` :P

src/wantlist/index.js
dignifiedquire: I'm okay with refactoring it so that it's not needed anymore, but there are places where changing this breaks things, that's all I wanted to point out.

test/wantmanager/index.spec.js
diasdavid: This changed to cid :)

dignifiedquire: good, because I think we are all out of `ew` buffers

src/components/decision-engine/index.js
```diff
@@ -70,8 +69,9 @@ class DecisionEngine {
         log('got task')
 
         pull(
-          this.blockstore.getStream(nextTask.entry.cid),
+          this.blockstore.getStream(nextTask.entry.cid.toV0()),
```
diasdavid: I'll change this one to `multihash`, as it might look confusing.

src/components/decision-engine/index.js
```diff
@@ -168,9 +167,8 @@ class DecisionEngine {
       ledger.wants(entry.cid, entry.priority)
 
       // If we already have the block, serve it
-      this.blockstore.has(entry.cid, (err, exists) => {
+      this.blockstore.has(entry.cid.toV0(), (err, exists) => {
```
diasdavid: and also here

src/components/decision-engine/index.js
dignifiedquire: done

src/components/decision-engine/index.js
dignifiedquire: done

diasdavid:
### Update 

This PR is pretty much in a `done` state. The only two tests missing are the deserialization and serialization of BitswapMessages that came from go, but that is (I believe) cause they are mangled, as we can use the Protobuf to serialize and deserialize fine.

Some remarks, Bitswap, as a thing, is really complex and it understandable that it requires a deep dive to understand, however, the code isn't really helping, I hope with this PR it welcomes more contributors as I tries to structure and refactor to minimize the cognitive overhead that one has to take in order to tinker with the code. Other than the things listed to improve in the first comment, here are some more vectors where we can improve:

- Less pull-streams overload, there were (and still are some cases) of pull-streams within pull-streams without any comments of what is intended, it takes some seconds to dial everything in.
- The fact that we use native Maps, means that we get a log of `.entries()`, which in Bitswap message, `entry` has a meaning. I wonder how we can make this more explicit, to avoid data type head 💥

@dignifiedquire thank you for adding a lot of tests to this module, it made the migration so much easier and I was very confident as I walked through it.

Tomorrow I will update the Block Service to use this new bitswap and then DAG API :)

diasdavid:
Now the code for this PR is present on: https://github.com/ipfs/js-ipfs-bitswap/pull/84

I've opened issues to handle all the concerns presented in this PR:

- https://github.com/ipfs/js-ipfs-bitswap/issues/85
- https://github.com/ipfs/js-ipfs-bitswap/issues/86
- https://github.com/ipfs/js-ipfs-bitswap/issues/87
- https://github.com/ipfs/js-ipfs-bitswap/issues/88
- https://github.com/ipfs/js-ipfs-bitswap/issues/89
- https://github.com/ipfs/js-ipfs-bitswap/issues/90
- https://github.com/ipfs/js-ipfs-bitswap/issues/91

PR #82, opened 2016-12-14
Improve perf by reducing calls to mh.toB58String

dignifiedquire:
We were abusing this encoding very heavily. Ref https://github.com/libp2p/js-peer-id/pull/44

PR #83, opened 2016-12-15
Improve performance

dignifiedquire:
- Drop peer request queue, we don't need it at this point
- reduce key to string conversions
- aggregate wantlist generation 
- properly cancel wants that we canceld
- extract benchmarks and expand them 

Includes #82


dignifiedquire:
There were issues with shutting down in our tests & benchmarks. It seems I finally got to the bottom of those:

- https://github.com/libp2p/js-libp2p-swarm/pull/139
- https://github.com/libp2p/js-libp2p-spdy/pull/47

now we don't need the timeout anymore `test/network/gen-network.node.js`

dignifiedquire:
Upgraded benchmarks are currently running locally:

```
  2 nodes - 1 blocks/node - 388ms
  2 nodes - 10 blocks/node - 554ms
  2 nodes - 100 blocks/node - 1478ms
  5 nodes - 1 blocks/node - 845ms
  5 nodes - 10 blocks/node - 3105ms
  5 nodes - 100 blocks/node - 13860ms
  10 nodes - 1 blocks/node - 3192ms
  10 nodes - 10 blocks/node - 36894ms
```

Where one block is `1024` bytes.

dignifiedquire:
Latest commit improves the benchmark to

```
2 nodes - 1 blocks/node - 353ms
  2 nodes - 10 blocks/node - 442ms
  2 nodes - 100 blocks/node - 437ms
  5 nodes - 1 blocks/node - 546ms
  5 nodes - 10 blocks/node - 661ms
  5 nodes - 100 blocks/node - 2455ms
  10 nodes - 1 blocks/node - 692ms
  10 nodes - 10 blocks/node - 1156ms
  10 nodes - 100 blocks/node - 49132ms
```

diasdavid:
One thing that we should do is make things like peerId.toB58String() become a cached property of the PeerId class itself, this way, instead of forcing the developer (in bitswap us, but for any app on top too) to remember to cache it always, V8 will start inlining every peerId.toB58String() as soon as the first time it is called. 

To save even more time, just make the PeerId class immutable all the way and cache all of these:
https://github.com/libp2p/js-peer-id/blob/master/src/index.js#L62-L68, then we can call toB58String or toJSON as many times we want and it will have 0 cost thanks to V8 inlining magic

dignifiedquire:
@diasdavid funny you mention that: https://github.com/libp2p/js-peer-id/pull/44

diasdavid:
@dignifiedquire PR's everywhere \o/ ! :D

dignifiedquire:
Closing in favor of #84

PR #113, opened 2017-02-19
feat(decision-engine): split large block messages

dignifiedquire:


dignifiedquire:
@VictorBjelkholm updated

src/components/decision-engine/index.js
```diff
@@ -35,14 +38,42 @@ class DecisionEngine {
   }
 
   _sendBlocks (env, cb) {
+    // split into messges of max 512 * 1024 bytes
```
whyrusleeping: Ah yes, because of https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/network/ipfs_impl.go#L186

We have a limit to the maximum message size you can send.

dignifiedquire: Yes, though that is higher than the multiplex limit you implemented in go. So I'm not sure if you could run into the same issue

PR #142, opened 2017-04-04
Peer and Content Routing (DHT)

dignifiedquire:
> WIP


src/index.js
```diff
@@ -198,7 +204,14 @@ class Bitswap {
...
 
       addListener()
-      this.wm.wantBlocks([cid])
+
+      this.network.findAndConnect(cid, CONSTANTS.maxProvidersPerRequest, (err) => {
+        if (err) {
+          return callback(err)
+        }
+
+        this.wm.wantBlocks([cid])
```
diasdavid: move wantBlocks out, it can be done in parallel

src/index.js
```diff
@@ -198,7 +204,14 @@ class Bitswap {
       }
 
       addListener()
-      this.wm.wantBlocks([cid])
+
+      this.network.findAndConnect(cid, CONSTANTS.maxProvidersPerRequest, (err) => {
```
diasdavid: We will need to avoid doing a DHT query for every block

src/components/network/index.js
```diff
@@ -99,24 +100,39 @@ class Network {
...
-    }
+    this.libp2p.dial(peerId, (err, conn) => {
+      if (err) {
+        return callback(err)
+      }
+
+      pull(pull.empty, conn)
+      callback()
+    })
+  }
```
diasdavid: Dialing without a protocol doesn't create a `conn` because it never opens a multiplex conn, it is what is called 'warming up a conn'.

This can be just: `this.libp2p.dial(peerId, callback)`

diasdavid:
Left: 
- [ ] update deps (use libp2p with DHT)
- [ ] Write tests for get and getMany
- [ ] Bubble up getMany to BlockService and Unixfs-Engine (separate PRs)

diasdavid:
This is ready for a round of review. 

I feel that network.js is wrapping libp2p APIs more than needed. I understand that the origin is because this module was a map from go-ipfs-bitswap and so it carries all of its weight when it comes to APIs, we can trim it though :)

diasdavid:
Quick realization. Bitswap needs to still work if no DHT is passed. Will update the PR with all the updated deps and make it work that way.

diasdavid:
To merge this PR, we need:

- [ ] Tests that test it with and without a DHT, so make sure it works in both cases.

test/bitswap.js
```diff
@@ -0,0 +1,149 @@
...
+        // TODO: on a cancel, should it yield a error saying it was cancelled?
+        expect(err).to.not.exist()
+        expect(block).to.not.exist()
+        console.log('--> I have concluded that I cant access the block')
+        // done()
+      })
+
+      // setTimeout(() => nodes[0].bitswap.unwant(block.cid), 2000)
+    })
+  })
```
diasdavid: I noticed something strange with this test. It seems that bitswap gives up too easily in trying to find a block. See the logs, even before it sends out the want list, it just returns.

```
» DEBUG=bitswap:* npm run test:node

> ipfs-bitswap@0.14.0 test:node /Users/koruza/code/js-ipfs-bitswap
> aegir-test node

[22:07:09] Starting 'test:node'...
[22:07:09] Starting 'mocha'...


  bitswap without DHT
    ✓ connect 0 -> 1 && 1 -> 2 (62ms)
Sat, 15 Jul 2017 21:07:10 GMT bitswap:error Failed to provide: DHT is not available
Sat, 15 Jul 2017 21:07:10 GMT bitswap:wantmanager adding to wl
--> I have concluded that I cant access the block
Sat, 15 Jul 2017 21:07:10 GMT bitswap:wantmanager:queue sending message
Sat, 15 Jul 2017 21:07:10 GMT bitswap:network sendMessage to QmcuySKjW2Xnsm9ETu2d8XAwxRgvqmYe26kv9akYEWcmuV BitswapMessage {
  full: false,
  wantlist:
   Map {
     '\u0012 M�\u001d����8�\u0019�\u0019��+I!\u001f1,<��\u0017]�t����' => BitswapMessageEntry { entry: [Object], cancel: false } },
  blocks: Map {} }
Sat, 15 Jul 2017 21:07:10 GMT bitswap:network incomming new bitswap connection: /ipfs/bitswap/1.1.0
Sat, 15 Jul 2017 21:07:10 GMT bitswap:network ending connection
```

diasdavid: @dignifiedquire thoughts?

src/network.js
```diff
@@ -0,0 +1,181 @@
...
+  connectTo (peer, callback) {
+    if (!this._running) { return callback(new Error(`network isn't running`)) }
+
+    this.libp2p.dial(peer, callback)
+  }
+
+  // Dial to the peer and try to use the most recent Bitswap
+  _dialPeer (peer, callback) {
+   // Attempt Bitswap 1.1.0
+    this.libp2p.dial(peer, BITSWAP110, (err, conn) => {
```
diasdavid: Removing the try/catch block here showed that our mocks were actually not correct. The try catch around these calls was making so that any error that would happen after a dial would end up here to be dealt from bitswap possibly causing a lot of other erros.

test/bitswap.js
diasdavid: I've detected this same problem in a couple of other tests now too.

test/bitswap.js
```diff
@@ -0,0 +1,146 @@
...
+      nodes = results
+      done()
+    })
+  })
+
+  after((done) => {
+    each(nodes, (node, cb) => {
+      series([
+        (cb) => {
+          node.bitswap.stop()
```
diasdavid: - [ ] need to update this one

test/decision-engine/decision-engine.js
```diff
@@ -0,0 +1,244 @@
...
+          .to.eql(sanfrancisco.peer.toHexString())
+
+        expect(sanfrancisco.engine.peers()).to.include(seattle.peer)
+
+        expect(seattle.engine.peers())
+          .to.include(sanfrancisco.peer)
+        done()
+      })
+    })
+  })
```
diasdavid: This test started failing with the move to async initiation

test/utils/mocks.js
```diff
@@ -44,12 +68,22 @@ exports.mockNetwork = (calls, done) => {
...
     },
-    start () {
+    start (callback) {
+      setImmediate(() => callback())
+    },
+    findAndConnect (cid, maxProviders, callback) {
+      setImmediate(() => callback())
+    },
+    provide (cid, callback) {
+      setImmediate(() => callback())
```
diasdavid: We were missing to call a bunch of these callbacks with just `setImmediate(() => callback)`

test/decision-engine/decision-engine.js
```diff
@@ -0,0 +1,246 @@
...
+      const m = new Message(true)
+      sanfrancisco.engine.messageSent(seattle.peer)
+
+      seattle.engine.messageReceived(sanfrancisco.peer, m, (err) => {
+        expect(err).to.not.exist()
+
+        // TODO why were we expecting for two different peers to have the 
+        // same hexString?
+        // expect(seattle.peer.toHexString())
+        //  .to.eql(sanfrancisco.peer.toHexString())
```
diasdavid: @dignifiedquire this assertion started failing when i moved to an async start up but now that I read it, why were we expecting to have different peers with the same Id?

test/components/decision-engine/index-test.js
```diff
@@ -1,252 +0,0 @@
...
-        const sanfrancisco = res[0]
-        const seattle = res[1]
-
-        const m = new Message(true)
-        sanfrancisco.engine.messageSent(seattle.peer)
-        seattle.engine.messageReceived(sanfrancisco.peer, m, (err) => {
-          expect(err).to.not.exist()
-
-          expect(seattle.peer.toHexString())
-            .to.not.eql(sanfrancisco.peer.toHexString())
```
diasdavid: AH! the not was deleted. my bad...

diasdavid:
A test with DHT keeps failing on Circle, not being able to reproduce it locally though:

image

diasdavid:
Ok, so figured it out that someone the `10 seconds` interval for broadcasting the wantlist is making this fail on circle, seems that Circle is obnoxiously slow

src/index.js
```diff
@@ -51,6 +49,7 @@ class Bitswap {
   _receiveMessage (peerId, incoming, callback) {
     this.engine.messageReceived(peerId, incoming, (err) => {
       if (err) {
+        // TODO: Q: why do we just log and not return here?
         log('failed to receive message', incoming)
       }
```
diasdavid: - [ ] Understand why this Error is just logged and why the rest of the function continues.

src/index.js
```diff
@@ -180,26 +223,62 @@ class Bitswap {
...
-      this.notifications.once(blockEvent, blockListeners[cidStr])
+      this.notifications.once(
+        unwantEvent(cidStr),
+        unwantListeners[cidStr]
+      )
+      this.notifications.once(
+        blockEvent(cidStr),
+        blockListeners[cidStr]
+      )
     }
```
diasdavid: - [ ] Q: Can we make this notification/tracking gets system be an external and tested module?

src/index.js
```diff
@@ -180,26 +223,62 @@ class Bitswap {
...
+          if (has) {
+            locals.push(cid)
+          } else {
+            missing.push(cid)
+          }
+          cb()
+        })
+      }, cb),
+      (cb) => {
+        if (missing.length > 0) {        // TODO: Q: why do we just log and not return here?
```
diasdavid: - [ ] This comment doesn't make sense anymore

src/network.js
```diff
@@ -0,0 +1,181 @@
...
+  constructor (libp2p, bitswap, options) {
+    options = options || {}
+    this.libp2p = libp2p
+    this.bitswap = bitswap
+    this.b100Only = options.b100Only || false
+
+    this._running = false
+
+    // TODO: move this up to libp2p-swarm
+    // this.libp2p.swarm.setMaxListeners(CONSTANTS.maxListeners)
```
diasdavid: - [ ] Make sure this happens in Swarm

src/network.js
```diff
@@ -0,0 +1,181 @@
...
+      }
+      // TODO: why doesn't the error get propageted back??
+      writeMessage(conn, serialized, (err) => {
+        if (err) { log(err) }
+      })
+      callback()
+    })
+  }
+
+  // TODO: if this method is just to call libp2p.dial, better call libp2p dial
```
diasdavid: It's fine to leave it here.

src/index.js
dignifiedquire: Why would that be beneficial? We can test it more in here if we want to.

dignifiedquire:
@diasdavid I can spend some time with this today

src/index.js
diasdavid: Exactly to add more tests :)

dignifiedquire:
> We're adding this to many repos now, is this something we officially decided to support or we still prefer npm over yarn?

I thought we had made that decision last time we talked about things, and that only node@4 support was blocking it, which we have since dropped

src/index.js
```diff
@@ -13,6 +12,7 @@ const CONSTANTS = require('./constants')
 const WantManager = require('./want-manager')
 const Network = require('./network')
 const DecisionEngine = require('./decision-engine')
+const Notifications = require('./notifications')
```
diasdavid: \o/

src/index.js
diasdavid: Glad you agreed https://github.com/ipfs/js-ipfs-bitswap/pull/142/files/d48c20e091c90c1abb6f42862564ace978d09afc..f634c3a69304d0f3b3ba93480686272cca3ba1b7#r135027986

diasdavid:
@dignifiedquire please add the yarn lock file to .gitignore.

dignifiedquire:
Update, fixed most issues, but it seems something in the connectivity is not working as expected, failing one test on CI. It seems`findProviders` doesn't actually do it's job properly currently, in some cases at least. Debugging on what is happening

src/decision-engine/index.js
```diff
@@ -12,17 +13,16 @@ const values = require('lodash.values')
...
 
-const log = debug('bitswap:engine')
-log.error = debug('bitswap:engine:error')
-
-const Message = require('../../types/message')
-const Wantlist = require('../../types/wantlist')
+const Message = require('../types/message')
+const Wantlist = require('../types/wantlist')
 const Ledger = require('./ledger')
+const logger = require('../utils').logger
```
diasdavid: 👍

src/decision-engine/index.js
```diff
@@ -108,7 +107,7 @@ class DecisionEngine {
           blocks: blockList
         }, (err) => {
           if (err) {
-            log.error('failed to send', err)
+            this._log.error('failed to send', err)
```
diasdavid: If it fails to send, why des `this.messageSent` get called anyway?

src/index.js
```diff
@@ -3,55 +3,57 @@
...
+    return this._libp2p.peerInfo
   }
 
   // handle messages received through the network
   _receiveMessage (peerId, incoming, callback) {
     this.engine.messageReceived(peerId, incoming, (err) => {
       if (err) {
-        log('failed to receive message', incoming)
+        // TODO: Q: why do we just log and not return here?
+        this._log('failed to receive message', incoming)
```
diasdavid: @dignifiedquire comments here?

test/bitswap.js
```diff
@@ -126,8 +126,8 @@ describe('bitswap with DHT', () => {
 
   it('connect 0 -> 1 && 1 -> 2', (done) => {
     parallel([
-      (cb) => nodes[0].libp2pNode.dial(nodes[1].libp2pNode.peerInfo, cb),
-      (cb) => nodes[1].libp2pNode.dial(nodes[2].libp2pNode.peerInfo, cb)
+      (cb) => nodes[2].libp2pNode.dial(nodes[1].libp2pNode.peerInfo, cb),
+      (cb) => nodes[1].libp2pNode.dial(nodes[0].libp2pNode.peerInfo, cb)
```
diasdavid: What's the difference? (not the title of the test needs to change too)

dignifiedquire: @diasdavid look at the commit message

diasdavid: "experiment" got it :)

src/index.js
dignifiedquire: The reason is that we still want to process the message as much as possible, even if there was a problem in the decision engine. (Technically that err is always undefined atm, but that could change later down the line)

test/bitswap.js
```diff
@@ -134,9 +134,8 @@ describe('bitswap with DHT', () => {
   it('put a block in 2, get it in 0', (done) => {
     waterfall([
       (cb) => makeBlock(cb),
-      (block, cb) => setTimeout(() => cb(null, block), 1000),
       (block, cb) => nodes[2].bitswap.put(block, () => cb(null, block)),
-      (block, cb) => setTimeout(() => cb(null, block), 1000),
+      (block, cb) => setTimeout(() => cb(null, block), 400),
       (block, cb) => nodes[0].bitswap.get(block.cid, (err, blockRetrieved) => {
```
diasdavid: How was this causing the issue? @dignifiedquire ?

dignifiedquire: The fix was adding the timeout, it seems that the propagation after the put takes more time on CI. I think this fix is "good enough" to move forward with this implementation, but we should work on setting up more integration style tests to make sure content is distributed well in the actual tests.

diasdavid: Seems that you actually reduced the timeout, not added one. What am I missing?

dignifiedquire: @diasdavid looka the full changes I made not just that commit

diasdavid: ah, I looked at the commit that said "find the source."

src/want-manager/index.js
```diff
@@ -111,24 +112,25 @@ module.exports = class WantManager {
...
-
-      this.peers.forEach((p) => {
-        p.addMessage(fullwantlist)
+      this.wantlist.forEach((entry) => {
+        fullwantlist.addEntry(entry.cid, entry.priority)
       })
-    }, 10 * 1000)
+
+      this.peers.forEach((p) => p.addMessage(fullwantlist))
+    }, 60 * 1000)
```
diasdavid: Do we really want to wait a minute to rebroadcast??

dignifiedquire: We already generate a lot of traffic, so I would rather go to a low value and increase it if we see issues with it

test/bitswap.js
dignifiedquire: yeah, debugging on ci eventually results in less than readable commits :D

src/want-manager/index.js
diasdavid: Does this affect the result in tests?

dignifiedquire: Not at all, all tests worked when I disabled these entirely.

.gitignore
```diff
@@ -35,3 +35,5 @@ test/test-repo-for*
 docs
 
 test/test-repo/datastore
+
+yarn.lock
```
diasdavid: This could be added next to the package-lock.json

README.md
```diff
@@ -73,24 +63,21 @@ See https://ipfs.github.io/js-ipfs-bitswap
...
+└─── types
+    ├── message            # (Type) message that is put in the wire
     │   ├── entry.js
     │   ├── index.js
     │   └── message.proto.js
-    └── wantlist            # (Type) track wanted blocks
+    └── wantlist           # (Type) track wanted blocks
         ├── entry.js
         └── index.js
 ```
```
diasdavid: A though. I've been the one documenting pretty much the whole Bitswap. I do like to document things and make them easier for other people to use and contribute (its like teaching), however, it is kind of hard to when I'm mostly the one pushing the docs, graphs and writing other ideas down. It would be very helpful when the way that things are designed change (or created), to get the contributor to document as well.

This tree is missing the new notifications.js and I'm sure a lot of other thoughts that went into this. Could you try to capture all of them in writing to the best of your ability so that future selfs or contributors don't have to hit a wall? Thank you :)

src/decision-engine/index.js
```diff
@@ -49,15 +49,26 @@ class DecisionEngine {
...
       size += b.data.byteLength
 
-      if (size >= MAX_MESSAGE_SIZE) {
+      if (size >= MAX_MESSAGE_SIZE ||
+          // need to ensure the last remaining items get sent
+          outstanding === 0) {
         const nextBatch = batch.slice()
         batch = []
-        this._sendSafeBlocks(env.peer, nextBatch, cb)
+        this._sendSafeBlocks(env.peer, nextBatch, (err) => {
```
diasdavid: Just noticed now that we have this `env` thing. I'm curious to learn why have you adopted this strategy, doesn't seem to match your love for things like Flow where every argument gets a type and every function a signature. Here it is impossible for the reader to figure out the amount of variables that can go into this `env` object.

src/decision-engine/index.js
```diff
@@ -49,15 +49,26 @@ class DecisionEngine {
...
+          if (err) {
+            this._log('sendblock error: %s', err.message)
+          }
+          // not returning the error, so we send as much as we can
+          // as otherwise `eachSeries` would cancel
```
diasdavid: ❤️ thank you for documenting :)

src/index.js
diasdavid: Could you add that as a comment replacing my `TODO: Q`?

src/index.js
```diff
@@ -150,55 +155,100 @@ class Bitswap {
...
-      addListener()
-      this.wm.wantBlocks([cid])
+      addListeners(missing)
+      this.wm.wantBlocks(missing)
+
+      this.network.findAndConnect(cids[0], CONSTANTS.maxProvidersPerRequest, (err) => {
+        if (err) {
+          this._log.error(err)
+        }
+      })
```
diasdavid: What about moving this constant to the network module itself, it isn't like the API is exposed and this call can be invoked with another value.

src/network.js
diasdavid: Could you open a issue for this on Swarm and delete this comment, please?

dignifiedquire:
@diasdavid on the swarm issue, shouldn't we keep the call around until it is actually moved?

src/decision-engine/index.js
dignifiedquire: I am not sure to be honest, probably just wanted to make it work

diasdavid:
@dignifiedquire its one line change in Swarm, open a PR :D

src/decision-engine/index.js
dignifiedquire: no more env :)

diasdavid:
There you go https://github.com/libp2p/js-libp2p-swarm/pull/226

dignifiedquire:
@diasdavid I think I fixed all things you mentioned, except for more docs. I'll think about what to write tomorrow and open a separate PR for that.

diasdavid:
Thank you :)

Seems that CI is unhappy now though

dignifiedquire:
@diasdavid I don't even:

screen shot 2017-08-24 at 23 36 12
screen shot 2017-08-24 at 23 36 18

test/bitswap-mock-internals.js
```diff
@@ -109,7 +109,8 @@ describe('bitswap with mocks', () => {
       })
     })
 
-    it('multi peer', (done) => {
+    it('multi peer', function (done) {
+      this.timeout(10 * 1000)
```
diasdavid: no more global timeouts? I love it!

test/bitswap-mock-internals.js
```diff
@@ -110,7 +110,7 @@ describe('bitswap with mocks', () => {
     })
 
     it('multi peer', function (done) {
-      this.timeout(10 * 1000)
+      this.timeout(40 * 1000)
```
diasdavid: next up is 41 :P

dignifiedquire:
@diasdavid this is entirely ridiculous at this point, tests are timing out randomly on travis as soon as I increase timeout on one another starts failing. I don't think we can realistically leave timeouts low and continue using travis.. also at this point the work has nothing to do anymore with the actual content of the PR this has been an issue on `master` for some time already

diasdavid:
Ok, waiting for a last round of CI.

README.md
diasdavid: @dignifiedquire never received an acknowledgment here. Could you share your feedback?

PR #152, opened 2017-09-27
feature: ability to override default bitswap max message size

ya7ya:
This PR is a part of the fix for the browser to browser file transfer issue in `IPFS` https://github.com/ipfs/js-ipfs/issues/1019 .

This edit allows an override of `MAX_MESSAGE_SIZE` so it can be used in the browser via WebRTC Transport. I been testing and so far 32KB seem to work in the browser without getting throttled.

License: MIT
Signed-off-by: Yahya <ya7yaz@gmail.com>


ya7ya:
hey @diasdavid , Can you take a look at this PR? 
This minor modification allows bitswap to change `MAX_MESSAGE_SIZE` so it can be used in the browser via WebRTC Transport. I been testing and so far 32KB seem to work in the browser without getting throttled.

Let me know what you think.

Stebalien:
@diasdavid shouldn't this be handled by the WebRTC transport?

diasdavid:
@ya7ya is this PR something you still think you can finish?

ya7ya:
@diasdavid Hey david, I completely forgot about this PR, I'm so sorry about this. I'll try to finalize this and add the required tests as soon as possible. 🙇

Stebalien:
There are a *lot* of changes here so it's a bit hard to tell what this is actually doing now but, if this is doing what the title says, this is very much the *wrong* fix. Is it really not possible (and much simpler) to have the WebRTC transport write at most 32KiB at a time?

ya7ya:
hey @Stebalien , I rebased this PR to master so that's why there are alotta changes.

> if this is doing what the title says, this is very much the wrong fix.

Yes. This does what the title says. And I'd be very interested in doing the _correct_ fix and rejecting this PR if it's possible to override this at webRTC transport level.

Stebalien:
It should be fairly easy to fix in the webRTC transport. Unfortunately, my understanding of js streams is rather limited (I generally avoid working on javascript).

ya7ya:
@Stebalien I'm gonna look into it. probably we can throttle the pull request `conn` in the Transport itself. I'm not 100% sure but I'll see what I can do. Thanks for the tip, I appreciate your input 👍 💯

dryajov:
@ya7ya I believe, we can use something like this to chunk the stream at the transport level - https://pull-stream.github.io/#pull-block.

alanshaw:
+1 this needs to be handled in the webrtc libp2p transport.

daviddias:
@vasco-santos, @jacobheun this PR suggests that the transports should get some benchmarking treatment.

There are two low hanging fruits to make things fast:
- The change that @ya7ya suggests for the WebRTC transport
- Creating a simple-peer module that doesn't rely on Readable Streams

achingbrain:
I don't think this PR is necessary any more, the upstream issue it's supposed to fix has been addressed by libp2p refactors.

Please feel free to reopen if it's still a problem.

PR #154, opened 2017-11-08
Windows interop

richardschneider:


richardschneider:
@diasdavid before merging, need PR https://github.com/libp2p/js-libp2p-kad-dht/pull/20 to be released.  Then update this `package.json`

test/bitswap-mock-internals.js
```diff
@@ -109,7 +109,8 @@ describe('bitswap with mocks', () => {
       })
     })
 
-    it('multi peer', function (done) {
+    // TODO: This test is just taking too long
+    it.skip('multi peer', function (done) {
```
diasdavid: That is not a valid excuse to skip a test. This Window PR should not be disabling tests to please the AppVeyor CI

richardschneider:
looks like `js-ipfs-repo` is misbehaving. I'll investigate.

```
Error: version mismatch: expected v6, found v5
      at get (node_modules\ipfs-repo\src\version.js:57:27)
      at store.get (node_modules\ipfs-repo\src\version.js:31:9)
      at node_modules\graceful-fs\graceful-fs.js:78:16
      at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:513:3)
```

Stebalien:
@richardschneider we recently bumped the repo version in go-ipfs and I believe that may have been done in js as well. @dryajov?

richardschneider:
Thanks @Stebalien Many of the JS tests have a repo as a test fixture.  does this mean we have to update all the test fixtures!

test/bitswap-mock-internals.js
richardschneider: This is not just pleasing appveyor.  Both travis and circle fail with this test enabled.

Stebalien:
Not sure. Honestly, I know almost nothing about js-ipfs (I also don't really see the point in repo compatibility but that's a different issue).

richardschneider:
See ipfs repo issue [#152](https://github.com/ipfs/js-ipfs-repo/issues/152)

richardschneider:
@diasdavid LGTM please review

dryajov:
@richardschneider I believe you are right, we might need to get the fixtures upgraded, but @diasdavid should be able to tell for sure.

richardschneider:
@dryajov This would be a big task.  I suggest that `ipfs-repo` should gracefully upgrade.

diasdavid:
@Stebalien we set ourselves to interop fully with go in as many points as possible so that a user would not have to think which IPFS implementation it is using. Unfortunately, go-ipfs does change the repo often and there are no specs or compliance tests we can use to keep it up and so we've separated paths for a while now. That to say we still keep the promise of trying to be compatible with go-ipfs as best we can.

In this case, all that was needed was just bumping the repo version. We use copies of go-ipfs created repos so that we can test our serializers and deserializers correctly.

diasdavid:
Don't really know what happened in one of the 4 jobs of appveyor

image

Seems that it is not an issue from this module. Merging :)

richardschneider:
Yes, I've seem this error now and then.  It only happens on a PR build.  I was looking into this afternoon and got side tracked by repo version number.

PR #155, opened 2017-11-08
add missing multicodec dependency

pgte:
, which is required [here](https://github.com/ipfs/js-ipfs-bitswap/blob/ca8fa722caf4747e7a4aa09102f0593c97d39538/src/types/message/index.js#L9).

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "lodash.sortby": "^4.7.0",
     "lodash.uniqwith": "^4.5.0",
     "lodash.values": "^4.3.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "lodash.sortby": "^4.7.0",
     "lodash.uniqwith": "^4.5.0",
     "lodash.values": "^4.3.0",
+    "multicodec": "^0.2.5",
```
diasdavid: Please use ~ for < 1.0.0 deps

PR #159, opened 2017-11-29
port go tests

pgte:
Addresses #9, porting [go-ipfs bitswap tests](https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/bitswap_test.go) into here.

Also, because of making it harder to test, this PR proposes a simplification of the `getMany` implementation. Besides being simpler to track the algorithm, it also guarantees the ordering of returns, i.e., the order of the CIDs passed in is the same order of the respective blocks getting out.
This has the downside of the result array having holes in it because of cancelations.
(If this last is a serious breaking change, I could always filter the output before returning).


diasdavid:
> Besides being simpler to track the algorithm, it also guarantees the ordering of returns, i.e., the order of the CIDs passed in is the same order of the respective blocks getting out.

Given that some blocks will be local and others remote, it is better to start returning as soon as possible. For testing, you can always sort it there.

pgte:
@diasdavid as far as I can tell, the previous algorithm was doing the same, no?

diasdavid:
@pgte you seem to be correct. This is a perf optimization that bitswap needs then.

diasdavid:
Before merging this one, could you make sure that this doesn't create any breaking change by going to js-ipfs, npm linking and running unit tests and interop tests?

pgte:
@diasdavid interop tests run ok.
Unit tests fail, but at the same 3 places they're failing when using ipfs-bitswap master (screenshots attached)

1__bash


1__bash

BTW, this PR is still not done: I was awaiting feedback and then got side-tracked. Back on this!

pgte:
@diasdavid except for the `rebroadcastTimeout` tests, all possible tests in https://github.com/ipfs/go-ipfs/blob/master/exchange/bitswap/bitswap_test.go were replicated.

pgte:
Also, depends on this for tests to run: https://github.com/libp2p/js-libp2p/pull/138

diasdavid:
Trying to release https://github.com/libp2p/js-libp2p/pull/138 but npm is not cooperating. Will try again later (or ping me on IRC when you are around)

diasdavid:
released. kicked off CI to start again :)

PR #162, opened 2018-01-22
fix: getMany: ensuring we set the want list

pgte:
... independently of having all blocks or not.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -243,6 +243,9 @@ class Bitswap {
           (has, cb) => {
             pendingStart--
             if (has) {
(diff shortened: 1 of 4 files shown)


diasdavid:
@pgte seems that some race condition might be happening. See Jenkins and Circle

pgte:
I don't see how this commit could have triggered a race condition. In this test, we're putting a block on node 1 before the test, and then getting it on node 2. The stats update occurs independently from the get, so the order of the sequence shouldn't matter.
I'll push a fix to this test.

Issue #165, opened 2018-02-01
per-peer stats


pgte:
We need these stats per peer because of https://github.com/ipfs/dynamic-data-and-capabilities/issues/3 and https://github.com/ipfs/js-ipfs/pull/1198.

hacdias:
For ipfs/js-ipfs#1198, I'll need per-peer and per-protocol stats, of, at least total bandwidth in and total bandwidth out.

pgte:
@hacdias can you define what you mean by "protocol"? Is it the bitswap protocol (1.0.0 or 1.1.0), the transport being used or the multiplexed protocol (DHT, bitswap, etc)?

(If it's the latest, we can't do this at the bitswap level).

hacdias:
@pgte protocol is the transport being used. Just opened an issue on libp2p: https://github.com/libp2p/js-libp2p/issues/157

diasdavid:
done with #166 , thanks @pgte :)

@hacdias anything else?

hacdias:
@diasdavid what I really need to implement the Bandwidth Stats on `js-ipfs` is libp2p/js-libp2p#157 because it should measure all of the in and out bandwidth and not only bitswap.

bvand:
@hacdias I'm new to ipfs, looking to help out on an issue like this to get started.

I was looking at https://github.com/ipfs/js-ipfs/pull/1198 - just to confirm my understanding, Bandwidth stats are the only additional per-peer stats left to implement?
What is the current status? Is this something that it would make sense for me to jump on?

bvand:
Found https://github.com/ipfs/js-ipfs/pull/1230 - looks like this is complete then? Or is there more?

PR #166, opened 2018-02-02
per-peer stats

pgte:
Addresses #165 

- [x] Starting off by documenting current stats API
- [x] Stats by peer

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,77 @@ Loading this module through a script tag will make the `IpfsBitswap` object avai
 
 See https://ipfs.github.io/js-ipfs-bitswap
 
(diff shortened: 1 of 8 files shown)


pgte:
I think this is all ready to review.
I'm not including the per-protocol stats because, according to #165, the per-protocol stats are only viewable at the global level, not at bitswap.

pgte:
@diasdavid it's the infamous "Multiplexer destroyed" error, which I think happens when reconnecting.

I've created #167 to address this.

diasdavid:
Ack. Let's move ahead then and figure out that one in parallel.

PR #175, opened 2018-05-18
fix: reset batch size counter

vmx:
When a batch of blocks was sent, the counter for the calculating
the size wasn't reset. After the initial batch, every message was
sent individually, which lead to a performance bottleneck. With
this change receiving larger files (> 2MiB) is about 30% faster.


dignifiedquire:
thank you for catching this @vmx!

PR #179, opened 2018-07-02
feat: change bitswapLedgerForPeer output format

wraithgar:
This matches the go implementation more closely


test/bitswap-mock-internals.js
```diff
@@ -78,11 +78,11 @@ describe('bitswap with mocks', function () {
...
-            expect(ledger.peer).to.equal(other.toB58String())
-            expect(ledger.value).to.equal(0)
-            expect(ledger.sent).to.equal(0)
-            expect(ledger.recv).to.equal(96)
-            expect(ledger.exchanged).to.equal(2)
+            expect(ledger.Peer).to.equal(other.toPrint())
+            expect(ledger.Value).to.equal(0)
+            expect(ledger.Sent).to.equal(0)
+            expect(ledger.Recv).to.equal(96)
+            expect(ledger.Exchanged).to.equal(2)
```
diasdavid: We just need to do this change at the http-api level to match go-ipfs. The internals should be doing the JS way.

vmx:
Do I understand @diasdavid correctly that this PR can be closed? If yes please close it. (The other thing in this PR was only about updating dependencies, which already happened).

diasdavid:
@vmx correct. Closing.

wraithgar:
I thing this should be reopened to at least switch to the `toPrint` method.

vmx:
@wraithgar Oh, i missed that toPrint() thing. Can you do another PR please?

wraithgar:
@diasdavid commented that it should not be changed but I disagree. You already have the real PeerId from when you passed it in to get this call, should you need it. It is quite clunky to convert this b58 encoded string into the results of `toPrint` down the wire.

vmx:
@wraithgar I think I agree with @diasdavid that this API should be JS friendly and things for the HTTP level are then made Go implementation compatible.

Aren't all b58 encoded CIDs formatted in a similar way in Go? A general `toPrint()` could then be re-used for other outputs.

wraithgar:
@vmx the `toPrint` for peerId is a custom string, not the b58 encoded CID.

vmx:
@wraithgar: Yes exactly. What i meant is: Doesn't Go always use the same kind of custom string whenever it outputs a CID?

wraithgar:
@vmx that seems irrelevant to this situation. The PeerId has a toPrint function already that matches the go implementation, it should be used.

Generalizing this would require that generic function being made aware of the type of object being represented, and feels a little too clever for its own sake.  Objects needing to conform to this format can just implement it.

vmx:
@wraithgar: thanks for the information. I haven't looked too closely. It surely doesn't make sense to be too clever.

PR #180, opened 2018-07-02
Use 1024 bit for testing + update deps (including libp2p)

wraithgar:
This speeds them up somewhat, and every little bit helps here when you're running them often.


vmx:
I forgot to mention that I normally let the author merge. So @wraithgar, feel free to push the button :)

vmx:
K, cool. Though I also wouldn't have minded merging this one and having a separate one for making the tests on Node.js 10 green.

test/utils/create-libp2p-node.js
```diff
@@ -8,33 +8,46 @@ const KadDHT = require('libp2p-kad-dht')
...
 class Node extends libp2p {
-  constructor (peerInfo, options) {
-    options = options || {}
-
-    const modules = {
-      transport: [new TCP()],
-      connection: {
-        muxer: Multiplex,
-        crypto: SECIO
+  constructor (_options) {
```
vmx: Why do you use an underscore in front of `options`. I normally use a leading underscore to signal that a variable isn't used (like Erlang and Rust do it).

diasdavid:
Chrome is just giving up in both Circle CI and Jenkins on the Linux host.

vmx:
A note for the future. Squashing to unrelated things like updating deps and changing some config value totally destroys the Git history. A clean history is useful for:

 - Understanding what a change was about in case the commit message isn't clear
 - Especially when upgrading deps, it's useful to know which changes were needed. So other modules can use that as a template to upgrade as well.
 - Being able to revert a single, self-contained thing, without reverting a lot of other unrelated things
 - Make it possible to find bugs easier when a `git bisect` is needed

test/utils/create-libp2p-node.js
```diff
@@ -8,33 +8,46 @@ const KadDHT = require('libp2p-kad-dht')
...
+      config: {
+        dht: {},
+        EXPERIMENTAL: {
+          dht: Boolean(_options.DHT)
+        }
+      }
     }
 
-    super(modules, peerInfo, null, options.DHT || {})
+    delete _options.DHT
```
VictorBjelkholm: I don't think it's a huge problem in this particular case but we should avoid using the `delete` operator on arguments to functions, as it'll change the `options` object that was passed into the constructor

PR #181, opened 2018-07-09
feat: change bitswapLedgerForPeer output format

wraithgar:
Use toPrint for the peer id output

As discussed in the last pr #179 we keep the lower case json output but switch to the correct peer id representation.

### DIFF
--- a/src/decision-engine/index.js
+++ b/src/decision-engine/index.js
@@ -143,7 +143,7 @@ class DecisionEngine {
       return null
     }
     return {
-      peer: ledger.partner.toB58String(),
+      peer: ledger.partner.toPrint(),
       value: ledger.debtRatio(),
       sent: ledger.accounting.bytesSent,
       recv: ledger.accounting.bytesRecv,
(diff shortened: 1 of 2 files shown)


vmx:
On #179 I agreed not being to clever about implementing things (having a general converter to print JS objects in a Go like way).

Though I'm still unsure if this change is right. Do I understand it correctly that the Go HTTP returns the peer in this format? But its only the HTTP, it's not the API itself. So I think the JS implementation should behave the same. The API should return the peer in the JS native format and the HTTP should then convert to whatever Go is doing to be compatible.

Coming back to not being too clever: This would been that you would just return the peer and the HTTP will call the `toPrint()` method.

wraithgar:
I didn't see that to go implementation differentiated its output based on the interface it was returning.  

The decision engine uses the `String` function for the partner (a peer id instance).  This is what formats the id in this `<Peer ...` output.

https://github.com/ipfs/go-ipfs/blob/f5d810932e871646c977c9aa51a10bcdbf58c4c9/exchange/bitswap/decision/engine.go#L122

vmx:
Thanks for pointing to the source code, you're absolutely right.

PR #183, opened 2018-12-05
Pass maxProviders to content routing

jacobheun:
With libp2p 0.24 you can now pass options objects to findProviders. The latest dht updates also include support for maxProviders.

This updates the call to `findProviders` so that both the maxProviders and maxTimeout options are sent.

Libp2p 0.24 will be in the js-ipfs 0.34 release, so we should be able to get this update in there as well.

**Also in this PR**:
* Update dependencies, also use libp2p-mplex instead of libp2p-multiplex
* Fix linting as a result of the latest Aegir
* Drop non jenkins ci files
* Remove .npmignore in favor of package.json `files` attribute
* Change tests to use 512 instead of 1024 keys, we dont need 1024 for testing

The bit change to the key reduced the `test:node` time locally quite a bit:
**Before**
```
real    0m39.634s
user    0m26.516s
sys     0m1.095s
```
**After**
```
real    0m24.447s
user    0m11.426s
sys     0m1.023s
```


package.json
```diff
@@ -86,8 +90,7 @@
     "varint-decoder": "~0.1.1"
   },
   "pre-push": [
-    "lint",
-    "test"
+    "lint"
```
vmx: Can we please keep the pre-push testing. I think they run fast enough so that it is OK.

jacobheun: Added that back.

PR #184, opened 2018-12-20
avoid running sync code in asynchronously

jacobheun:
This PR refactors some of the bitswap code to prevent it from running synchronous code asynchronously, which can bloat the call stack and in severe cases contribute to stack overflows. 

This also changes usage of setImmediate to use nextTick instead, as there is a performance boost there.

Also:
* Fixed an issue where the want manager tests could call done multiple times
* The decision engine had a waterfall callback that was never being called, that's been fixed. 

Relates to https://github.com/libp2p/js-libp2p-switch/issues/287


src/index.js
```diff
@@ -146,7 +147,7 @@ class Bitswap {
   _putBlock (block, callback) {
     this.blockstore.put(block, (err) => {
       if (err) {
-        return callback(err)
+        return setImmediate(() => callback(err))
```
vmx: I tried to learn about `setImmediate()` and found https://stackoverflow.com/questions/47184855/when-is-setimmediate-required-in-node-callback

If that description is accurate, you wouldn't need a `setImmedate()` here as the callback comes from the outer function.

But I certainly lack a good understanding, so I might as well be wrong.

For the same cases I'll just comment "not sure if needed".

src/index.js
```diff
@@ -333,7 +334,7 @@ class Bitswap {
       }, cb),
       (newBlocks, cb) => this.blockstore.putMany(newBlocks, (err) => {
         if (err) {
-          return cb(err)
+          return setImmediate(() => cb(err))
```
vmx: not sure if needed

src/index.js
```diff
@@ -345,7 +346,7 @@ class Bitswap {
             }
           })
         })
-        cb()
+        setImmediate(cb)
```
vmx: not sure if needed

test/wantmanager/index.spec.js
```diff
@@ -63,14 +63,16 @@ describe('WantManager', () => {
         expect(calls.connects).to.have.length(6)
         expect(calls.messages).to.have.length(6)
 
-        calls.messages.forEach((m, i) => {
-          expect(m[0]).to.be.eql(calls.connects[i])
+        for (let i = 0; i < calls.messages.length; i++) {
+          const m = calls.messages[i]
```
vmx: Just a note (no need to change it). I generally try not to use single letter variables (for easier grepping). A nice trick a colleague showed me as to simply double the letter if you want a single letter variable. So it would be `ii` and `mm`.

src/types/message/index.js
```diff
@@ -187,7 +188,7 @@ BitswapMessage.deserialize = (raw, callback) => {
   if (decoded.payload.length > 0) {
     return each(decoded.payload, (p, cb) => {
       if (!p.prefix || !p.data) {
-        cb()
+        return nextTick(cb)
```
alanshaw: ooh good catch!

src/index.js
```diff
@@ -108,7 +109,7 @@ class Bitswap {
       (has, cb) => {
         this._updateReceiveCounters(peerId.toB58String(), block, has)
         if (has) {
-          return cb()
+          return nextTick(cb)
```
alanshaw: In this case you're already on a future tick from the `blockstore.has` call above.

src/index.js
```diff
@@ -310,7 +311,7 @@ class Bitswap {
       (cb) => this.blockstore.has(block.cid, cb),
       (has, cb) => {
         if (has) {
-          return cb()
+          return nextTick(cb)
```
alanshaw: Same here, `blockstore.has` puts you on a future tick.

src/index.js
jacobheun: I'd recommend we leave these in place to avoid the internals here being sync _or_ async depending on the existence of the block. We should try to keep the function footprint the same imo.

alanshaw:
@vmx can we get this merged and released please 🙏

PR #187, opened 2019-01-09
fix: stringify to hex

### LINKED ISSUE #186: Non unique string CID keys
As noted here https://github.com/libp2p/js-libp2p-floodsub/pull/59, two buffers of differing content do not necessarily produce a unique utf8 string when `toString` is called.

I noticed `toString` is being called on CID buffers, and that string is being used as a key for the CID that is wanted, if the buffer doesn't make a unique string then bitswap could potentially signal that we have a CID that we don't have!

https://github.com/ipfs/js-ipfs-bitswap/blob/399dc0ed09a36dae6c5df741eb59cc9026b2c473/src/notifications.js#L52

vmx:
Calling `toString()` on a buffer might lead to the same result
although the buffers are different. Example:

    $ node
    > const buf1 = Buffer.from('15603533e990dfde', 'hex')
    undefined
    > const buf2 = Buffer.from('15603533e990dfe0', 'hex')
    undefined
    > buf1.toString()
    '\u0015`53���'
    > buf2.toString()
    '\u0015`53���'
    > buf1.toString() === buf2.toString()
    true
    > buf1.toString('hex') === buf2.toString('hex')
    false

Currently CIDs are stored internally with their `toString()` representation.
This could fail as different CIDs could end up with the same representation.
The fix is to use the hex representation instead.

Closes #186.

BREAKING CHANGE: emitted keys change

Bitswap has an event emitter which emits events when a block is received
and when a block is not wanted anymore. These events contain a string
representation of the CID. This representation changes with this commit.

Though practically it shouldn't have any impact on any system, except if
you have hard-coded specific CIDs.

### DIFF
--- a/src/notifications.js
+++ b/src/notifications.js
@@ -34,7 +34,7 @@ class Notifications extends EventEmitter {
    * @return {void}
    */
   hasBlock (block) {
-    const str = `block:${block.cid.buffer.toString()}`
+    const str = `block:${block.cid.buffer.toString('hex')}`
     this._log(str)
     this.emit(str, block)
   }
@@ -49,7 +49,7 @@ class Notifications extends EventEmitter {
    * @returns {void}
    */
   wantBlock (cid, onBlock, onUnwant) {
-    const cidStr = cid.buffer.toString()
+    const cidStr = cid.buffer.toString('hex')
     this._log(`wantBlock:${cidStr}`)
 
     this._unwantListeners[cidStr] = () => {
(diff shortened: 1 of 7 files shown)


alanshaw:
I'd like to know what the impact of this is - is IPFS using the events or are they just used internally in this module?

vmx:
That's a very good question. I have no clue who is using those events, they are not used (AFAIK) internally.

daviddias:
@alanshaw @vmx can we get this one merged and then create a PR to remove these?

vmx:
> create a PR to remove these?

"these" the events? If yes, then I'd remove the events first. What would be the point of breaking them first and then removing them? :)

daviddias:
Fair point. Have you tried removing them?

vmx:
Superseded by https://github.com/ipfs/js-ipfs-bitswap/pull/191.

PR #191, opened 2019-04-08
fix: base encode CIDs before logging or emitting them

achingbrain:
At the moment we turn the CID buffers into UTF8 encoded strings before logging them and emitting events with that string in the name.  This can result in Zalgo leaking into the terminal:

```console
bitswap:notif:QmZoFh73 wantBlock: å&���+�
                                           2�@��Hfä4���┼�▒3��]� +0└⎽
  ␉␋├⎽┬▒⎻:┬▒┼├:Q└Z⎺F␤73 ▒␍␍␋┼± ├⎺ ┬┌ +0└⎽
  ┌␋␉⎻2⎻:␍␤├:Q└Z⎺F␤73 °␋┼␍P⎼⎺┴␋␍␊⎼⎽ Q└␉1≥Q⎺7RK▒␍±␌8␍4RFFBVP┘134N␌2│FEPP5┬S␋M␊GAAT4 +1└⎽
  ┌␋␉⎻2⎻:␍␤├:⎻⎼⎺┴␋␍␊⎼⎽:Q└Z⎺F␤73 ±␊├P⎼⎺┴␋␍␊⎼⎽ Q└␉1≥Q⎺7RK▒␍±␌8␍4RFFBVP┘134N␌2│FEPP5┬S␋M␊GAAT4 +1└⎽
  ␉␋├⎽┬▒⎻:┼⎺├␋°:Q└Z⎺F␤73 ┬▒┼├B┌⎺␌┐: �◆+�W���␌��#�����␋��4�Ɔ┌# +1└⎽
  ␉␋├⎽┬▒⎻:┬▒┼├:Q└Z⎺F␤73 ▒␍␍␋┼± ├⎺ ┬┌ +1└⎽
  ┌␋␉⎻2⎻:␍␤├:Q└Z⎺F␤73 °␋┼␍P⎼⎺┴␋␍␊⎼⎽ Q└U4␌K▒Q␊┤R␌│YK┬HYFSGZCF≤B1─␊┤L␌X┴Y┘┬┼┐AUT└KAS +0└⎽
  ┌␋␉⎻2⎻:␍␤├:⎻⎼⎺┴␋␍␊⎼⎽:Q└Z⎺F␤73 ±␊├P⎼⎺┴␋␍␊⎼⎽ 
```

This PR calls `cid.toBaseEncodedString()` instead of `cid.buffer.toString()`.

May be a breaking change as the names of events emitted by this module are based on `cid.buffer.toString()`.

### DIFF
--- a/src/notifications.js
+++ b/src/notifications.js
@@ -34,7 +34,7 @@ class Notifications extends EventEmitter {
    * @return {void}
    */
   hasBlock (block) {
-    const str = `block:${block.cid.buffer.toString()}`
+    const str = `block:${block.cid}`
     this._log(str)
     this.emit(str, block)
   }
@@ -49,7 +49,7 @@ class Notifications extends EventEmitter {
    * @returns {void}
    */
   wantBlock (cid, onBlock, onUnwant) {
-    const cidStr = cid.buffer.toString()
+    const cidStr = cid.toString()
     this._log(`wantBlock:${cidStr}`)
 
     this._unwantListeners[cidStr] = () => {
@@ -80,7 +80,7 @@ class Notifications extends EventEmitter {
    * @returns {void}
    */
   unwantBlock (cid) {
-    const str = `unwant:${cid.buffer.toString()}`
+    const str = `unwant:${cid}`
     this._log(str)
     this.emit(str)
   }
--- a/test/notifications.spec.js
+++ b/test/notifications.spec.js
@@ -34,7 +34,7 @@ describe('Notifications', () => {
   it('hasBlock', (done) => {
     const n = new Notifications(peerId)
     const b = blocks[0]
-    n.once(`block:${b.cid.buffer.toString()}`, (block) => {
+    n.once(`block:${b.cid}`, (block) => {
       expect(b).to.eql(block)
       done()
     })


vmx:
There's a similar fix at https://github.com/ipfs/js-ipfs-bitswap/pull/187.

@achingbrain Is it urgent? If not, I'd like to remove the notification stuff first (https://github.com/ipfs/js-ipfs-bitswap/pull/187#issuecomment-462404533).

achingbrain:
It will make debugging things a lot more pleasant as your terminal output wouldn't occasionally turn into gibberish.

achingbrain:
Could this be merged please?  Or #187.  Either way.

So much Klingon in my terminal.

image

achingbrain:
Actually with `cids@0.7.0` and later `cid.toString()` returns the CID base encoded so we can just call that.

vmx:
@achingbrain I've updated the commit message with a breaking change notice. If you approve I'll merge and do a release.

achingbrain:
Not sure I can approve my own PR in the GH sense, but:

LGTM 👍

achingbrain:
Should really be an `(un)BREAKING CHANGE:` as the CID buffers we were previously encoding as text can have unprintable characters which lead to clashes where the wrong event handlers are called as we fixed in a different PR in a different repo that I can't find right now. 😂

alanshaw:
I'm just looking at the changes here and wondering about the feature of CIDs where they [remember their multibase](https://github.com/multiformats/js-cid#cidtobaseencodedstringbasethismultibasename).

For example, we could ask for a CIDv1 that was encoded as base64 originally. So calling `toString()` would result in a base64 encoded CID and when we finally get the block we get a CIDv1 which will default to base32 when `toString()` is called (since CID is communicated as buffer with no multibase) - these won't match and we won't resolve the want.

@achingbrain @dirkmc

dirkmc:
That's a good point @alanshaw

@achingbrain I guess we should be explicit about which base we are converting to when converting to string

alanshaw:
Added an issue to track this https://github.com/ipfs/js-ipfs-bitswap/issues/200

PR #192, opened 2019-04-14
feat: use async/await

nijynot:
Early draft of making `js-ipfs-bitswap` async/await or AsyncIterable.

* `get()` now has an option whether to prompt the network or not, defaults to `true` which is similar to previous behaviour.
* Using the same `extendIterator` from `js-ipld`.


dirkmc:
Thanks @nijynot!

Closing in favour of https://github.com/ipfs/js-ipfs-bitswap/pull/202 that is a little further along (but still needs lots of work)

PR #197, opened 2019-06-14
perf: use async/setImmediate instead of async/nextTick

### LINKED ISSUE #196: Use async/setImmediate instead of async/nextTick
I've noticed that this module is using `async/nextTick` instead of `async/setImmediate`. Using `async/setImmediate` is preferred due to the way `async/nextTick` works in the browser:

> Calls callback on a later loop around the event loop. In Node.js this just calls process.nextTick. In the browser it will use setImmediate if available, otherwise setTimeout(callback, 0), which means other higher priority events may precede the execution of callback.

Because [timers are often throttled](https://developers.google.com/web/updates/2017/03/background_tabs) when a tab is not active to improve battery life, they only run at 1Hz (once every second). This has a drastic performance impact in those scenarios.

Will you accept a PR to change this?

Meanwhile, I'm aliasing `async/nextTick` to `async/setImmediate` in my webpack config.

satazor:
Building tools such as Webpack, automatically inject process.nextTick which uses https://github.com/defunctzombie/node-process/blob/master/browser.js, which relies on timers.
Because it uses timers, things become very slow when the tab is not focused/visible because browsers throttle timers (1Hz), see https://developers.google.com/web/updates/2017/03/background_tabs.

The setImmediate polyfill uses a postMessage trick to make it work reliable, even if the tab is not focused.

Related: https://github.com/webtorrent/webtorrent/issues/1568

Closes #196

### DIFF
--- a/src/decision-engine/index.js
+++ b/src/decision-engine/index.js
@@ -3,7 +3,7 @@
 const each = require('async/each')
 const eachSeries = require('async/eachSeries')
 const waterfall = require('async/waterfall')
-const nextTick = require('async/nextTick')
+const setImmediate = require('async/setImmediate')
 
 const map = require('async/map')
 const debounce = require('just-debounce-it')
@@ -66,7 +66,7 @@ class DecisionEngine {
           cb()
         })
       } else {
-        nextTick(cb)
+        setImmediate(cb)
       }
     }, cb)
   }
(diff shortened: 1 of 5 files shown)


dirkmc:
Closing for the reasons outlined in https://github.com/ipfs/js-ipfs-bitswap/issues/196

PR #199, opened 2019-06-17
feat: use ipfs provider

vasco-santos:
BREAKING CHANGE: ipfs provider was added in constructor.

In the context of `js-ipfs` reannouncing blocks to the network [ipfs/js-ipfs#2160](https://github.com/ipfs/js-ipfs/issues/2160), this PR aims to extract the provide logic from `bitswap`, allowing `js-ipfs` to be able to control how and when the provide/reprovide happens

This is part of the work for [ipfs/js-ipfs#2184](https://github.com/ipfs/js-ipfs/pull/2184), and that PR should be reviewed before merging this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "peer-book": "~0.9.0",
     "peer-id": "~0.12.0",
     "peer-info": "~0.15.0",
+    "promisify-es6": "^1.0.3",
     "rimraf": "^2.6.2",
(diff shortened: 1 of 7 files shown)


SgtPooki:
I think since js-ipfs is deprecated we can close this, but we may want to remove the ipfs provider stuff if it hasn't been already.

cc @achingbrain

PR #202, opened 2019-07-01
Feat/async await

dirkmc:
Supersedes https://github.com/ipfs/js-ipfs-bitswap/pull/192

This PR depends on a number of async/await PRs that are still WIP (peer-id, peer-info, etc) so it is still some way from being ready.


src/decision-engine/index.js
```diff
@@ -90,35 +81,25 @@ class DecisionEngine {
...
-          cb()
-        })
-      }, callback)
-    ], (err) => {
-      this._tasks = []
-
-      if (err) {
-        this._log.error(err)
+    const blocks = await Promise.all(uniqCids.map(cid => this.blockstore.get(cid)))
+    await Object.values(groupedTasks).map(async (tasks) => {
```
hugomrdias: This doesnt need a Promise.all to be awaited?

src/index.js
```diff
@@ -224,61 +196,46 @@ class Bitswap {
    * @param {function(Error, Blocks)} callback
```
hugomrdias: remove this line

src/index.js
```diff
@@ -307,51 +264,43 @@ class Bitswap {
    * @param {function(Error)} callback
```
hugomrdias: remove this line

src/index.js
```diff
@@ -307,51 +264,43 @@ class Bitswap {
...
+    return this.putMany(block)
   }
 
   /**
    * Put the given blocks to the underlying blockstore and
    * send it to nodes that have it them their wantlist.
    *
-   * @param {Array<Block>} blocks
+   * @param {AsyncIterable<Block>} blocks
    * @param {function(Error)} callback
```
hugomrdias: remove this line

src/index.js
```diff
@@ -307,51 +264,43 @@ class Bitswap {
...
-    ], callback)
+    return this.putMany(block)
   }
 
   /**
    * Put the given blocks to the underlying blockstore and
    * send it to nodes that have it them their wantlist.
    *
-   * @param {Array<Block>} blocks
+   * @param {AsyncIterable<Block>} blocks
```
hugomrdias:
```suggestion
   * @param {AsyncIterable<Block>|Iterable<Block>} blocks
```

src/network.js
```diff
@@ -67,13 +60,12 @@ class Network {
     pull(
       conn,
       lp.decode(),
-      pull.asyncMap((data, cb) => Message.deserialize(data, cb)),
+      pull.asyncMap((data, cb) => Message.deserialize(data).then(data => cb(null, data), cb)),
```
hugomrdias: call `cb` inside setImmediate when mixing promise and callbacks

src/network.js
```diff
@@ -67,13 +60,12 @@ class Network {
...
       pull.asyncMap((msg, cb) => {
         conn.getPeerInfo((err, peerInfo) => {
           if (err) { return cb(err) }
 
-          // this._log('data from', peerInfo.id.toB58String())
-          this.bitswap._receiveMessage(peerInfo.id, msg, cb)
+          this.bitswap._receiveMessage(peerInfo.id, msg).then(() => cb(), cb)
```
hugomrdias: call `cb` inside setImmediate when mixing promise and callbacks

src/network.js
```diff
@@ -97,102 +89,120 @@ class Network {
...
+    await this.libp2p.contentRouting.provide(cid)
   }
 
   // Connect to the given peer
   // Send the given msg (instance of Message) to the given peer
-  sendMessage (peer, msg, callback) {
-    if (!this._running) { return callback(new Error(`network isn't running`)) }
+  // sendMessage (peer, msg, callback) {
+  async sendMessage (peer, msg) {
+    // if (!this._running) { return callback(new Error(`network isn't running`)) }
```
hugomrdias: remove comment

src/types/message/index.js
```diff
@@ -41,7 +39,7 @@ class BitswapMessage {
   }
 
   addBlock (block) {
-    assert(Block.isBlock(block), 'must be a valid cid')
+    assert(Block.isBlock(block), 'must be a valid block')
```
hugomrdias: Can we remove `assert` here ? it's big for the browser and it doesn't add much value.

src/utils.js
```diff
@@ -80,11 +80,41 @@ const sortBy = (fn, list) => {
...
+  }
+  return values
+}
+
+const extendIterator = (iterator) => {
+  iterator.first = () => first(iterator)
+  iterator.last = () => last(iterator)
+  iterator.all = () => all(iterator)
+  return iterator
+}
```
hugomrdias: i wouldn't extend an iterator, would just use the helpers above

src/decision-engine/index.js
achingbrain: `this.blockstore.get(cid)` returns a promise, if that's what you mean?

hugomrdias: No Promise.all(Object.values().map() )

src/network.js
achingbrain: Why? As I understand it the function passed to `.then` is already going to be deferred to the microtask queue at end of the event loop, `setImmediate` will execute it on the next loop.

src/utils.js
achingbrain: I don't think this is even used.

src/types/message/index.js
achingbrain: Death to types.

src/index.js
achingbrain: Comments are not a type system. 😉

src/decision-engine/index.js
achingbrain: 😬

src/index.js
hugomrdias: they can be (ts can check types with jsdoc) and editor autocomplete is underrated (for me is very helpful)

src/network.js
hugomrdias: i have answered this question many times so this time a made an example i can point to. 
https://codesandbox.io/s/mixing-promises-with-callbacks-i1w0z

hope this explains the problem.

achingbrain: Gotcha.  Have updated.  We'll pull all that out once libp2p ships it's pull-streams --> async/interator refactor anyway.

src/index.js
```diff
@@ -71,53 +65,46 @@ class Bitswap {
...
-        if (has || !wasWanted) {
-          return nextTick(cb)
-        }
+    const has = await this.blockstore.has(block.cid)
+    this._updateReceiveCounters(peerId.toB58String(), block, has)
 
-        this._putBlock(block, cb)
-      }
-    ], callback)
+    if (!wasWanted) {
```
dirkmc:
```suggestion
    if (has || !wasWanted) {
```

src/network.js
```diff
@@ -97,102 +92,118 @@ class Network {
...
-          callback(null, conn, BITSWAP100)
-        })
-
-        return
+  async _dialPeer (peer) {
+    try {
+      // Attempt Bitswap 1.1.0
+      return {
+        conn: await this.libp2p.dialProtocol(peer, BITSWAP110),
+        protocol: BITSWAP100
```
dirkmc:
```suggestion
        protocol: BITSWAP110
```

src/network.js
```diff
@@ -97,102 +92,118 @@ class Network {
...
       }
-
-      callback(null, conn, BITSWAP110)
-    })
+    } catch (err) {
+      // Attempt Bitswap 1.0.0
+      return {
+        conn: await this.libp2p.dialProtocol(peer, BITSWAP100),
+        protocol: BITSWAP110
```
dirkmc:
```suggestion
        protocol: BITSWAP100
```

dirkmc:
It makes me a bit nervous that the tests were passing when the protocols were switched around. @achingbrain do you think it's worth adding a test to catch that case?

achingbrain:
It worked because all nodes in the tests are JS nodes and the message deserialization code [handles both message versions](https://github.com/ipfs/js-ipfs-bitswap/blob/facaa87a2c9900d8f5b92056d329c3025a2d5301/src/types/message/index.js#L146-L173) by checking to see if the relevant fields exist.

I guess we could add a 'strict' mode that only processes one version of bitswap but that sounds beyond the scope of this PR.

dirkmc:
Ok fair enough - I can't officially approve as I opened the PR, but LGTM 👍 

Let me know if and when you want me to merge + release

achingbrain:
Yes please, whenever you can :)

PR #203, opened 2019-10-19
fix: reduce size

hugomrdias:
change lodash.isequalwith for a custom much smaller function specific for this use case.

72.22KB > 67.87KB gzipped

Probably much faster also.

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,7 +1,7 @@
 'use strict'
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


dirkmc:
Nice, thank you

Could you please add a test for the new function 🙏

hugomrdias:
@dirkmc should be ready to merge

dirkmc:
Thanks 🎉

PR #204, opened 2019-12-11
Support responding to new message types

dirkmc:
We are currently [extending go-bitswap](https://github.com/ipfs/go-bitswap/pull/189) with some new functionality.

To help Bitswap to more efficiently direct requests and utilize bandwidth, the response will include `PendingBytes`: the number of bytes that the responder has queued up to send to the requestor (blocks are sent out on a round robin basis, so if a node receives a lot of requests it may build up a backlog).

To help disseminate information about block distribution across the swarm there are some new message types:
- Requests
  - *want-have*
    Tell me if you have a block (but don't send it)
  - *want-block*
    Send me the block if you have it
- Responses
  - HAVE
    I have the block
  - DONT_HAVE
    I don't have the block. Note that this message is sent immediately, and is only sent if the requestor asks for an immediate response.

This branch serves as the baseline for the server side changes. Smaller PRs will be merged into this branch.
- [x] update message protobuf
- [x] Support `PendingBytes` https://github.com/ipfs/js-ipfs-bitswap/pull/205
- [x] Support HAVE / DONT_HAVE https://github.com/ipfs/js-ipfs-bitswap/pull/211
- [x] Ensure incoming blocks are added to blockstore before being passed to engine https://github.com/ipfs/js-ipfs-bitswap/pull/211
- [x] Listen on bitswap protocol 1.2.0 https://github.com/ipfs/js-ipfs-bitswap/pull/212


src/types/message/message.proto.js
```diff
@@ -5,11 +5,18 @@
...
+        Have = 1;
+      }
+
       message Entry {
         // changed from string to bytes, it makes a difference in JavaScript
         optional bytes block = 1;      // the block cid (cidV0 in bitswap 1.0.0, cidV1 in bitswap 1.1.0)
         optional int32 priority = 2;    // the priority (normalized). default to 1
         optional bool cancel = 3;       // whether this revokes an entry
+        WantType want_type = 4;         // Note: defaults to enum 0, ie Block
+        bool send_dont_have = 5;        // Note: defaults to false
```
alanshaw: Is there _any_ way we can have `wantType`, `sendDontHave` etc.? 🙏 Snake case is not idiomatic JS and we already have to interact with Pascal case because of golang. It would be good not to add another! 😇

dirkmc: hmm yeah I'm not sure why I made those snakey in the first place 🐍

achingbrain: FWIW the protocol buffers style guide recommends snake_case for property names and CamelCase for message names.

https://developers.google.com/protocol-buffers/docs/style

dirkmc:
Interop tests for file transfer appear to be comparable for most cases, with a significant speedup for fetching directories from JS.

Note: master on the left, this branch on the right:

results

PR #205, opened 2019-12-11
Include pending bytes in response message

dirkmc:
See https://github.com/ipfs/js-ipfs-bitswap/pull/204

### DIFF
--- a/src/decision-engine/index.js
+++ b/src/decision-engine/index.js
@@ -28,15 +28,17 @@ class DecisionEngine {
 
   async _sendBlocks (peer, blocks) {
     // split into messages of max 512 * 1024 bytes
-    const total = blocks.reduce((acc, b) => {
+    let total = blocks.reduce((acc, b) => {
       return acc + b.data.byteLength
     }, 0)
 
+    // If the blocks fit into one message, send the message right away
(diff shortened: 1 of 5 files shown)


src/decision-engine/index.js
```diff
@@ -45,6 +47,7 @@ class DecisionEngine {
       outstanding--
       batch.push(b)
       size += b.data.byteLength
+      total -= b.data.byteLength
 
       if (size >= MAX_MESSAGE_SIZE ||
```
alanshaw: Is there a bug here? We add to the batch until size is greater than (or equal to) `MAX_MESSAGE_SIZE` and then we send that batch...surely that message is then too big to send?

dirkmc: That's a good question - `MAX_MESSAGE_SIZE` is set to 256k, but in reality we should be able to send messages an order of magnitude larger than that. So I think probably this is by design: we want to send _at least_ MAX_MESSAGE_SIZE and a little bit more. This is actually how go-bitswap does it.

I'm currently working on a PR to change how all this works so my inclination is to ignore for this PR and fix in my next PR, does that sound reasonable?

alanshaw: 👍 SGTM

PR #206, opened 2019-12-12
refactor: update network to use async libp2p

jacobheun:
Migrating to async libp2p. 

**Note**: This is currently using the prerelease

- [x] Depends on [libp2p 0.27](https://github.com/libp2p/js-libp2p/issues/487)


jacobheun:
This is still pending the release of libp2p 0.27, but should otherwise be good to review. I refactored some things that were related to the libp2p change, but tried to avoid touching too much. I wouldn't mind helping do some more refactoring as a followup PR (after the new year).

src/network.js
```diff
@@ -17,78 +16,80 @@ class Network {
...
-          callbackify(this.bitswap._receiveMessage.bind(this.bitswap))(peerInfo.id, msg, cb)
-        })
-      }),
-      pull.onEnd((err) => {
-        this._log('ending connection')
-        if (err) {
-          this.bitswap._receiveError(err)
         }
-      })
+      }
```
alanshaw: try/catch around this pipeline? I think the remote can send a stream reset message that will throw.

src/network.js
```diff
@@ -130,10 +134,10 @@ class Network {
   async sendMessage (peer, msg) {
     if (!this._running) throw new Error('network isn\'t running')
 
-    const stringId = peer.toB58String() ? peer.toB58String() : peer.id.toB58String()
+    const stringId = peer.toB58String()
     this._log('sendMessage to %s', stringId, msg)
 
-    const { conn, protocol } = await this._dialPeer(peer)
+    const { stream, protocol } = await this._dialPeer(peer)
```
alanshaw: Does this need a try/catch?

src/network.js
```diff
@@ -17,78 +16,80 @@ class Network {
...
-        conn.getPeerInfo((err, peerInfo) => {
-          if (err) {
-            return cb(err)
+      async (source) => {
+        for await (const data of source) {
+          try {
+            const message = await Message.deserialize(data.slice())
+            this.bitswap._receiveMessage(connection.remotePeer, message)
+          } catch (err) {
+            this.bitswap._receiveError(err)
```
alanshaw: Do we want to continue receiving or return?

jacobheun: To be the same as the existing code it should end iteration, you're right. I added a `break` to exit iteration to keep it consistent with the existing functionality.

src/network.js
jacobheun: Yes, good catch, added.

src/network.js
jacobheun: It shouldn't need one because `sendMessage` already throws. There are catches in both places sendMessage is used.

dirkmc:
@jacobheun please let me know when libp2p is released and I will update the dependency and merge

jacobheun:
@dirkmc will do, we just have two outstanding docs PRs, https://github.com/libp2p/js-libp2p/pull/514 and https://github.com/libp2p/js-libp2p/pull/508. Once those are merged I will do an RC and follow up shortly after that with a release (potentially tomorrow, pending reviews).

dirkmc:
Cool thank you, no rush ❤️

jacobheun:
@dirkmc I've updated libp2p, this should be all set.

PR #209, opened 2019-12-22
fix: use multicodec correctly

hacdias:
Correctly uses `multicodec` API instead of going around! Please see https://github.com/multiformats/js-multicodec/issues/54.

License: MIT
Signed-off-by: Henrique Dias <hacdias@gmail.com>


alanshaw:
@dirkmc please can this be released asap - this has broken all js-ipfs installs.

dirkmc:
Thanks for the fix @hacdias

dirkmc:
Landed in v0.26.2

PR #210, opened 2019-12-23
chore: update multicodec

hacdias:
Updating multicodec to 1.0.0 as we decided to bump it to 1.0.0 and deprecate 0.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "ipfs-block": "~0.8.0",
     "just-debounce-it": "^1.1.0",
     "moving-average": "^1.0.0",
-    "multicodec": "~0.5.7",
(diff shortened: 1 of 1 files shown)


adamsoffer:
@hacdias this commit seems to have caused a compilation error in typescript applications that depend on this library.

image

I believe this is because it's referring to a file that's been removed in the latest multicodec release (`name-table.js`)?

adamsoffer:
ah looks like this was addressed here https://github.com/ipfs/js-ipfs-bitswap/pull/209

hacdias:
@adamsoffer yes, and only Windows is failing, as on master. Merry Christmas too!

achingbrain:
This was done as part of #206

PR #212, opened 2020-01-10
Listen on bitswap/1.2.0

dirkmc:
We need to receive messages on bitswap protocol 1.2.0 as part of https://github.com/ipfs/js-ipfs-bitswap/pull/204 (respond to want-have / want-block messages)

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -10,6 +10,7 @@ const logger = require('./utils').logger
 
 const BITSWAP100 = '/ipfs/bitswap/1.0.0'
 const BITSWAP110 = '/ipfs/bitswap/1.1.0'
+const BITSWAP120 = '/ipfs/bitswap/1.2.0'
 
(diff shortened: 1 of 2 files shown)


dirkmc:
The plan is to merge this into #204, I just broke #204 up into several different PRs to make it easier to review.

We need to listen on bitswap 1.2.0 because the go-ipfs PoC will [attempt to connect on](https://github.com/ipfs/go-bitswap/pull/189/files#diff-33bf651b39540d23d5b24eb4a54026cdR111) bitswap 1.2.0, and if so it will [send the new message types](https://github.com/ipfs/go-bitswap/pull/189/files#diff-6d9ddbabb69a60a8680f10e99fb14e4eR375) (want-have / want-block). If not it will revert to using the old message types (regular want).

Does that sound like the right approach?

PR #215, opened 2020-03-04
fix: blocks may not be in the blockstore yet

achingbrain:
Because some blockstore implementations batch up write before committing them to disk or wherever, they may not be available when we think they should be.

I noticed this module throwing unhandledPromiseRejections around this so the change here handles missing blocks by putting the task back into the queue to process later, with a fairly arbitrary 5x limit on processing a given task.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "promisify-es6": "^1.0.3",
     "rimraf": "^3.0.0",
     "safe-buffer": "^5.1.2",
+    "sinon": "^9.0.0",
     "stats-lite": "^2.2.0",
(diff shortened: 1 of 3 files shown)


dirkmc:
The way this works has changed with https://github.com/ipfs/js-ipfs-bitswap/pull/211, which is close to being merged. My preference would be to concentrate on getting that PR merged and then tackle this issue.

achingbrain:
I think the code can still cause `UnhandledPromiseRejection`s but the implementation has changed.  I will address it in a subsequent PR.

PR #216, opened 2020-03-05
feat: use bitswap fetch queue

dirkmc:
Allow multiple concurrent requests for the same block


achingbrain:
My feeling is that this module has too many places that hold state, so any solution should not add new places for state to hide.

PR #217, opened 2020-04-21
feat: use libp2p 0.28.x

vasco-santos:
Once `js-libp2p@0.28` ships, we will have to update its breaking changes in here and `js-ipfs`.

For what is relevant for this module, we are deprecating `peer-info`, changing how `js-libp2p` gathers its listening multiaddr, changing the `ConnectionManager`, `PeerStore` and `DHT` APIs.

Needs:

- [x]  [libp2p/js-libp2p#610](https://github.com/libp2p/js-libp2p/pull/610)
- [x]  [libp2p/js-libp2p#611](https://github.com/libp2p/js-libp2p/pull/611) -- This PR needs to use the new `ConnectionManager` API instead of Registrar for getting connections
- [x] [libp2p/js-libp2p-kad-dht#181](https://github.com/libp2p/js-libp2p-kad-dht/pull/181)
- [x] #214 + rebase
- [x]  needs `libp2p@0.28` final release

BREAKING CHANGE: bitswap start and stop return a promise


vasco-santos:
@dirkmc @achingbrain this is ready for review. Once this and [ipfs/js-ipfs#3019](https://github.com/ipfs/js-ipfs/pull/3019) is good to go, we should release the final version of `libp2p`

package.json
```diff
@@ -92,8 +90,7 @@
     "varint-decoder": "^0.4.0"
   },
   "pre-push": [
-    "lint",
-    "test"
+    "lint"
```
achingbrain, line 93: Did you mean to remove the test run?

test/network/gen-bitswap-network.node.js
```diff
@@ -79,7 +79,7 @@ async function exchangeBlocks (nodes, blocksPerNode = 10) {
   const d = Date.now()
 
   // fetch all blocks on every node
-  await Promise.all(nodes.map(async (node) => {
+  await Promise.all(nodes.map(async (node, index) => {
```
achingbrain, line 82: `index` is unused?

vasco-santos:
FYI I addressed the review and we are now only pending on the final release of `js-libp2p`. We will do it, once the `js-ipfs` integration PR is reviewed, so that we are comfortable with the final release.

I will let you know once that happens

vasco-santos:
@dirkmc @achingbrain 

I was experiencing intermittent issues using this in `js-ipfs`.

Let's consider the following scenario:

- PeerB wants a block from the network
- PeerA dials PeerB's multiaddr
  - a connection is established between them
  - PeerA knows one of the PeerB's listen multiaddrs, since it was used for dial
  - PeerB still does not know any PeerA's listen multiaddr, since it was dialled.
- PeerB is notified by the connectionManager of the new connection by its event
  - PeerB tries to dial PeerA to get the intended block, but it does not have multiaddrs for it in the AddressBook
  - Dial fails and PeerB will not be able to ask PeerA for the block
- (meanwhile) The identify protocol finishes and PeerB now knows PeerA listen multiaddrs

The issue here was that PeerB got notified of the connection between both peers and it should be notified only when that peer is dialable on bitswap protocol.

I needed to change that flow to use the [libp2p/js-libp2p-interfaces//topology](https://github.com/libp2p/js-libp2p-interfaces/tree/master/src/topology). This will trigger the onConnect when the identify protocol ran and the peer knowns that the other peer is running one of the multicodecs of bitswap. By this time, the peer already added the other peer multiaddrs to the AddressBook and this intermittent issue does not happen.

It is important pointing out that this is a breaking change in the bitswap API, since start and stop will now return Promise

src/network.js
```diff
@@ -33,29 +35,36 @@ class Network {
...
-    this.libp2p.on('peer:disconnect', this._onPeerDisconnect)
+    // register protocol with topology
+    const topology = new MulticodecTopology({
+      multicodecs: this.protocols,
+      handlers: {
+        onConnect: this._onPeerConnect,
+        onDisconnect: this._onPeerDisconnect
+      }
+    })
+    this._registrarId = await this.libp2p.registrar.register(topology)
```
achingbrain, line 50: Does this need to be async?  The `Registrar.register` method doesn't appear to be an async method: https://github.com/libp2p/js-libp2p/blob/master/src/registrar.js#L64-L78

vasco-santos: lol, yes my bad! 😓We had it async before, but not anymore. I will change this according

dirkmc:
Thanks for the update vasco, I appreciate that you're doing such thorough testing that you found this 👍

src/network.js
vasco-santos: just amended the previous commit removing the async part, as well as the breaking change message. Thanks for noticing this

test/network/network.node.js
```diff
@@ -134,7 +136,8 @@ describe('network', () => {
 
       bitswapMockB._receiveError = (err) => deferred.reject(err)
 
-      const { stream } = await p2pA.dialProtocol(p2pB.peerInfo, '/ipfs/bitswap/' + version.num)
+      // TODO: set addr
```
jacobheun, line 139: @vasco-santos the address should be added. The previous test is causing this to pass but if that changes this will break

vasco-santos:
final comments were addressed in the last commit

PR #218, opened 2020-04-23
fix: use ipld-block and remove node globals

hugomrdias:
- add buffer to the deps
- swaps ipfs-block with ipld-block
- removes safe-buffer
- updates deps

BREAKING CHANGE: swaps ipfs-block with ipld-block

related to https://github.com/ipfs/js-ipfs/issues/2924


dirkmc:
Thanks @hugomrdias ❤️

PR #224, opened 2020-05-26
perf: decrease wantlist send debounce time

achingbrain:
We wait 200ms before sending wantlists to remote peers which is an eternity. I'm not sure how this figure was arrived at but I've reduced it to 10ms which has increased the js->js transfer times in the interop tests to:

| Data    | 200ms    | 10ms    | Speedup |
|---------|----------|---------|---------|
| 1.02 kB | 486ms    | 142ms   | 342%    |
| 63.5 kB | 449ms    | 119ms   | 377%    |
| 65.5 kB | 436ms    | 105ms   | 415%    |
| 524 kB  | 1287ms   | 210ms   | 613%    |
| 786 kB  | 1676ms   | 282ms   | 594%    |
| 1.05 MB | 2066ms   | 326ms   | 634%    |
| 1.05 MB | 2123ms   | 330ms   | 643%    |
| 4.19 MB | 7091ms   | 1045ms  | 679%    |
| 8.39 MB | 13711ms  | 1927ms  | 712%    |
| 67.1 MB | 107704ms | 13462ms | 800%    |
| 134 MB  | 214988ms | 26038ms | 826%    |

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -9,5 +9,6 @@ module.exports = {
   provideTimeout: 15 * SECOND,
   kMaxPriority: Math.pow(2, 31) - 1,
   rebroadcastDelay: 10 * SECOND,
-  maxListeners: 1000
+  maxListeners: 1000,
+  wantlistSendDebounceAmount: 10
 }
--- a/src/want-manager/msg-queue.js
+++ b/src/want-manager/msg-queue.js
@@ -4,6 +4,7 @@ const debounce = require('just-debounce-it')
 
 const Message = require('../types/message')
 const logger = require('../utils').logger
+const { wantlistSendDebounceAmount } = require('../constants')
 
 module.exports = class MsgQueue {
   constructor (selfPeerId, otherPeerId, network) {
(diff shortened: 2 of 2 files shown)


achingbrain:
Running with different debounce times and counting the network packets transmitted between nodes (pinning is disabled so the overall time is a lot less):

| Debounce    | Packets    | Time    | Speedup |
|---------|----------|---------|---------|
| 200ms | ~650    | 1266ms   | 0    |
| 10ms | ~700    | 301ms   | 420%    |
| - | ~750    | 252ms   | 502%    |

By comparison, go-ipfs@0.5.1 running the same test:

| Packets    | Time    |
|----------|---------|
| ~1300    | 269ms   |

Based on this we should probably disable the debounce entirely.

src/constants.js
```diff
@@ -9,5 +9,6 @@ module.exports = {
   provideTimeout: 15 * SECOND,
   kMaxPriority: Math.pow(2, 31) - 1,
   rebroadcastDelay: 10 * SECOND,
-  maxListeners: 1000
+  maxListeners: 1000,
+  wantlistSendDebounceAmount: 10
```
dirkmc, line 12: Nit: can we include the unit in the variable name, eg `wantlistSendDebounceMs`

achingbrain: Sure, but none of the other have units.

PR #234, opened 2020-07-20
fix: pass peer id to onPeerConnect

achingbrain:
The persistent peer store in libp2p 0.28 means sometimes we know about peers when we start up, so we try to start bitswap for them when we start up if we have an open connection already.

The connection object has a remotePeer property that should be used instead of passing the whole thing to `this._onPeerConnect`.

Adds a test for the above.

Fixes ipfs/js-ipfs#3182

### DIFF
--- a/test/network/network.node.js
+++ b/test/network/network.node.js
@@ -11,67 +11,56 @@ const makeBlock = require('../utils/make-block')
 const Network = require('../../src/network')
 const Message = require('../../src/types/message')
 
+function createBitswapMock () {
+  return {
+    _receiveMessage: async () => {},
+    _receiveError: async () => {},
+    _onPeerConnected: async () => {},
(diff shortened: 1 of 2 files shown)


test/network/network.node.js
```diff
@@ -109,6 +98,38 @@ describe('network', () => {
...
+
+    bitswapMockB._onPeerConnected = (peerId) => {
+      expect(peerId.toB58String()).to.equal(p2pA.peerId.toB58String())
+      counter++
+    }
+
+    const ma = `${p2pB.multiaddrs[0]}/p2p/${p2pB.peerId.toB58String()}`
+    await p2pA.dial(ma)
+
+    await pWaitFor(() => counter >= 2)
```
vasco-santos, line 117: I would use `p-defer` instead of the counter

PR #261, opened 2020-10-14
feat: typedef generation & type checking

Gozala:
This pull request makes following changes:

1. Adds typedef generation
2. Adds github action to run typecheck on changes & report missmatches in pull requests
3. Adds / fixes bunch of jsdoc comments so it can both typecheck & generate

While making these changes I did discover certain things that are either bugs or I am misunderstanding what is going on there. I'll comment inline.

Please not that this change depends on:

1. https://github.com/pgte/moving-average/pull/9
2. https://github.com/ipfs/js-ipfs/pull/3580 (without this type check will fail)


src/decision-engine/index.js
```diff
@@ -112,7 +123,7 @@ class DecisionEngine {
 
     // If there's nothing in the message, bail out
     if (msg.empty) {
-      this._requestQueue.tasksDone(peerId, tasks)
+      peerId && this._requestQueue.tasksDone(peerId, tasks)
```
Gozala, line 126: `peerId` is obtained on line 76 via following code:

```js
 const { peerId, tasks, pendingSize } = this._requestQueue.popTasks(this._opts.targetMessageSize)
```

However `popTasks` returns an object without `peerId` in 2 code paths out of 3. At the same time `tasksDone` and `messageSent` below do not expect `undefined`.

It was not exactly clear what made most sense here, so I just conditioned those calls on `peerId`s presence, but it would be good to make sure all this makes sense.

src/decision-engine/req-queue.js
```diff
@@ -254,21 +274,31 @@ class PeerTasks {
...
   }
 
   /**
    * No work to be done, this PeerTasks object can be freed.
+   *
    * @returns {boolean}
    */
   isIdle () {
-    return this._pending.length === 0 && this._active.length === 0
+    return this._pending.length === 0 && this._active.size === 0
```
Gozala, line 291: It looks like there was a bug in the code, because `_active` is a `Set` instance which doesn't have `length` property.

src/stats/stat.js
```diff
@@ -118,18 +156,22 @@ class Stats extends EventEmitter {
...
+   * @private
+   * @param {Op} op
+   */
   _applyOp (op) {
     const key = op[0]
     const inc = op[1]
 
     if (typeof inc !== 'number') {
-      throw new Error('invalid increment number:', inc)
+      throw new Error(`invalid increment number: ${inc}`)
```
Gozala, line 168: This seemed like a bug, so I just inlined `inc`into error message.

src/types/message/index.js
```diff
@@ -154,11 +194,18 @@ class BitswapMessage {
...
+   * @returns {boolean}
+   */
   equals (other) {
     if (this.full !== other.full ||
         this.pendingBytes !== other.pendingBytes ||
         !isMapEqual(this.wantlist, other.wantlist) ||
         !isMapEqual(this.blocks, other.blocks) ||
+        // @TODO - Is this a bug ? `isMapEqual` only compare maps of blocks and
+        // wants
+        // @ts-ignore
```
Gozala, line 208: This one seems like a bug, but maybe it is not 🙄 So `isMapEqual` will compare only that keys match and will compare values if they are either `Block`s or `BitswapMessageEntry`s (anything that has `equals` method on them actually). However `blockPresences`  contain numbers so two would be equal even if the actual values are different, which doesn't seems  like intended behavior.

P.S.: TS not happy mostly because it's told to expect maps of blocks or entries and is passed maps of numbers below.

src/want-manager/msg-queue.js
```diff
@@ -7,13 +7,19 @@ const logger = require('../utils').logger
...
+   * @param {Network} network
+   */
   constructor (selfPeerId, otherPeerId, network) {
     this.peerId = otherPeerId
     this.network = network
     this.refcnt = 1
 
     this._entries = []
-    this._log = logger(selfPeerId, 'msgqueue', otherPeerId.toB58String().slice(0, 8))
+    this._log = logger(selfPeerId, 'msgqueue')
```
Gozala, line 22: Not sure what to do with the last argument here, because `logger` seems to take just two. I dropped it to retain the current behavior, but maybe concatenating last two would make more sense.

.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
```
hugomrdias, line 1: please use the package.json property for this

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,27 @@
+on:
```
hugomrdias, line 1: can you please use the aegir template https://github.com/ipfs/aegir/wiki/Github-Actions-Setup

package.json
```diff
@@ -8,6 +8,13 @@
     "./test/utils/create-libp2p-node": false,
     "./test/utils/create-temp-repo-nodejs.js": "./test/utils/create-temp-repo-browser.js"
   },
+  "typesVersions": {
```
hugomrdias, line 11: please follow aegir guidelines https://github.com/ipfs/aegir/blob/master/md/ts-jsdoc.md

package.json
```diff
@@ -17,14 +24,18 @@
     "test:browser": "aegir test -t browser -t webworker",
     "test:node": "aegir test -t node",
     "lint": "aegir lint",
+    "check": "tsc --noEmit",
```
hugomrdias, line 27:
```suggestion
    "check": "aegir ts -p check",
```

package.json
```diff
@@ -17,14 +24,18 @@
...
     "bench": "node benchmarks/index",
-    "build": "aegir build",
+    "build": "npm run build:js & npm run build:types",
+    "build:js": "aegir build",
+    "build:types": "tsc --emitDeclarationOnly --declarationDir dist",
     "coverage": "aegir coverage --provider codecov",
     "docs": "aegir docs",
-    "benchmarks": "node test/benchmarks/get-many"
+    "benchmarks": "node test/benchmarks/get-many",
+    "prepare": "npm run build:types"
```
hugomrdias, line 38: aegir build already does all of this

package.json
```diff
@@ -69,15 +80,17 @@
     "rimraf": "^3.0.0",
     "sinon": "^9.0.0",
     "stats-lite": "^2.2.0",
-    "uuid": "^8.0.0"
+    "uuid": "^8.0.0",
+    "typescript": "^4.0.3",
```
hugomrdias, line 84:
```suggestion
```
this comes from aegir

src/decision-engine/interface.ts
```diff
@@ -0,0 +1,51 @@
+export interface TaskMerger {
+  /**
+   * Ggiven the existing tasks with the same topic, does the task add some new
```
hugomrdias, line 3:
```suggestion
   * Given the existing tasks with the same topic, does the task add some new
```

.github/workflows/typecheck.yml
hugomrdias: Probably remove travis ci also

src/index.js
```diff
@@ -243,10 +295,10 @@ class Bitswap {
...
    * blockstore they are returned, otherwise the blocks are added to the wantlist and returned once another node sends them to us.
    *
-   * @param {AsyncIterator<CID>} cids
-   * @param {Object} options
-   * @param {AbortSignal} options.abortSignal
-   * @returns {Promise<AsyncIterator<Block>>}
+   * @param {AsyncIterable<CID>|Iterable<CID>} cids
+   * @param {Object} [options]
+   * @param {AbortSignal} [options.signal]
+   * @returns {AsyncIterable<Block>}
```
hugomrdias, line 301: shouldnt this be AsyncGenerator ?

src/index.js
```diff
@@ -305,7 +352,7 @@ class Bitswap {
    * Put the given blocks to the underlying blockstore and
    * send it to nodes that have it them their wantlist.
    *
-   * @param {AsyncIterable<Block>} blocks
+   * @param {AsyncIterable<Block>|Iterable<Block>} blocks
    * @returns {AsyncIterable<Block>}
```
hugomrdias, line 356:
```suggestion
   * @returns {AsyncGenerator<Block>}
```

src/stats/index.js
```diff
@@ -1,17 +1,28 @@
...
-const EventEmitter = require('events')
+const { EventEmitter } = require('events')
 const Stat = require('./stat')
 
+/**
+ * @typedef {[number, number, number]} AvarageIntervals
+ */
 const defaultOptions = {
-  movingAverageIntervals: [
+  movingAverageIntervals: /** @type {AvarageIntervals} */ ([
```
hugomrdias, line 10:
```suggestion
  movingAverageIntervals: /** @type {AverageIntervals} */ ([
```

src/stats/index.js
```diff
@@ -1,17 +1,28 @@
 'use strict'
 
-const EventEmitter = require('events')
+const { EventEmitter } = require('events')
 const Stat = require('./stat')
 
+/**
+ * @typedef {[number, number, number]} AvarageIntervals
```
hugomrdias, line 7:
```suggestion
 * @typedef {[number, number, number]} AverageIntervals
```

src/stats/stat.js
```diff
@@ -1,25 +1,38 @@
...
 
 class Stats extends EventEmitter {
+  /**
+   *
+   * @param {string[]} initialCounters
+   * @param {Object} options
+   * @param {boolean} options.enabled
+   * @param {number} options.computeThrottleTimeout
+   * @param {number} options.computeThrottleMaxQueueSize
+   * @param {import('.').AvarageIntervals} options.movingAverageIntervals
```
hugomrdias, line 15:
```suggestion
   * @param {import('.').AverageIntervals} options.movingAverageIntervals
```

please use full path and extract type to types.ts

src/types.ts
```diff
@@ -0,0 +1,64 @@
+import CID from 'cids'
+import PeerId from 'peer-id'
```
hugomrdias, line 2: i think theres a PeerId type being defined somewhere else in these changes.

tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
```
hugomrdias, line 1: use aegir template
`aegir ts -p config > tsconfig.json`

hugomrdias (changes requested):
can we make an issue for the all ts-ignore's  ?

package.json
```diff
@@ -69,15 +80,17 @@
     "rimraf": "^3.0.0",
     "sinon": "^9.0.0",
     "stats-lite": "^2.2.0",
-    "uuid": "^8.0.0"
+    "uuid": "^8.0.0",
```
Gozala, line 83:
```suggestion
    "uuid": "^8.0.0"
```

.github/workflows/typecheck.yml
Gozala: Can we please do the travis to github migration as a separate pr ?

src/index.js
Gozala: I think this is a big question which I boils down to:

1. Are we committed to supporting an AsyncGenerator API or not ?
2. Should change argument types to `AsyncGenerator`s as well ?
    _I do not see much value in making return types `AsyngGenerators` otherwise  as type checker will treat them as`AsyncIterable`s anyway.

Whatever we decide we should probably apply across the code base

Gozala: Oh there is more things to consider, type here is the TS interface for `AsyncGenerator`

```ts
interface AsyncGenerator<T = unknown, TReturn = any, TNext = unknown> extends AsyncIterator<T, TReturn, TNext> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...args: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
    return(value: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
    throw(e: any): Promise<IteratorResult<T, TReturn>>;
    [Symbol.asyncIterator](): AsyncGenerator<T, TReturn, TNext>;
}

type IteratorResult<T, TReturn = any> = IteratorYieldResult<T> | IteratorReturnResult<TReturn>;

interface IteratorYieldResult<TYield> {
    done?: false;
    value: TYield;
}

interface IteratorReturnResult<TReturn> {
    done: true;
    value: TReturn;
}

```

which raises following questions:

1. What should be the type for `TReturn`, specifically it the output position ? Probably `void`.

1. What should be a type for `TNext` & should we even accept arguments to `.next()` ?

1. And then there is the is still this loose `.throw`. What's the contract there ?

So if anything we should probably have a type alias that encodes some of the answers here e.g. `type Stream<T> = AsyncGenerator<T, void, void>`.

src/types.ts
Gozala: it is imported in couple of files where they are used

Gozala:
@hugomrdias I have addressed all of your feedback except:

1. Migration from travis to github actions. I don't think it makes sense to bundle it with changes here & separate PR would make more sense IMO.
2. `AsyncIterable` vs `AsyncGenerator` seems like a broader discussion to have.

Gozala:
Merging master introduced the problem with libp2p typedefs so temporarily this pull depends on https://github.com/libp2p/js-libp2p-interfaces/pull/73

Gozala:
> can we make an issue for the all ts-ignore's ?

https://github.com/ipfs/js-ipfs-bitswap/issues/273

Gozala:
@hugomrdias any idea what is going on here ? 
https://travis-ci.com/github/ipfs/js-ipfs-bitswap/jobs/449964224

For some reason `typescript` dependency does not get hoisted in the CI, while it does seem to get hoisted on my machine.

hugomrdias:
dunno maybe because some deps arent using aegir ts yet or the eslint config

package.json
```diff
@@ -69,19 +82,21 @@
...
+    "@types/debug": "^4.1.5",
+    "typescript": "^4.0.5"
```
hugomrdias, line 87:
```suggestion
```

hugomrdias (approved):

package.json
```diff
@@ -8,20 +8,33 @@
...
+  },
+  "eslintConfig": {
+    "extends": "ipfs"
+  },
   "files": [
     "dist",
     "src"
   ],
   "scripts": {
+    "prepare": "aegir build",
```
Gozala, line 28: We don't need full build here

```suggestion
    "prepare": "aegir ts -p check",
```

src/network.js
```diff
@@ -102,27 +132,38 @@ class Network {
...
+   * @param {Object} [options]
+   * @param {AbortSignal} [options.signal]
+   * @returns {AsyncIterable<Provider>}
    */
   findProviders (cid, maxProviders, options = {}) {
-    return this.libp2p.contentRouting.findProviders(
+    return this._libp2p.contentRouting.findProviders(
       cid,
       {
+        // TODO: Should this be a timeout options insetad ?
```
Gozala, line 165: @vasco-santos do you know if we should be passing `timeout` instead of `maxTimeout`

src/network.js
```diff
@@ -191,24 +240,36 @@ class Network {
...
+   * @param {AbortSignal} [options.signal]
    * @returns {Promise<Connection>}
    */
   async connectTo (peer, options) { // eslint-disable-line require-await
     if (!this._running) {
       throw new Error('network isn\'t running')
     }
 
-    return this.libp2p.dial(peer, options)
+    // TODO: Figure out inconsistency here.
```
Gozala, line 253: @vasco-santos could you please help me out understand what is expected here. Passed `peer` here comes from `libp2p.contentRouting.findProviders` which contains both peer id and multiaddr, but then `libp2p.dial` expects either `peerId` or `multiaddr` and attempts to turn it back into `Provider`.

I do not know all the ins and outs here, but clearly something makes wrong assumptions.

src/types/message/index.js
```diff
@@ -154,11 +195,17 @@ class BitswapMessage {
...
+   * @param {BitswapMessage} other
+   * @returns {boolean}
+   */
   equals (other) {
     if (this.full !== other.full ||
         this.pendingBytes !== other.pendingBytes ||
         !isMapEqual(this.wantlist, other.wantlist) ||
         !isMapEqual(this.blocks, other.blocks) ||
+        // @TODO - Is this a bug ?
+        // @ts-expect-error - isMap equals map values to be objects not numbers
```
Gozala, line 208: @hugomrdias This looks like a bug to me, could you please confirm or tell me if I'm not seeing something ?

src/want-manager/index.js
```diff
@@ -27,12 +45,14 @@ module.exports = class WantManager {
...
-          this.wantlist.removeForce(e.cid)
+          this.wantlist.removeForce(e.cid.toString())
         } else {
           this.wantlist.remove(e.cid)
         }
       } else {
         this._log('adding to wl')
+        // TODO: Figure out the wantType
+        // @ts-expect-error - requires wantType
         this.wantlist.add(e.cid, e.priority)
```
Gozala, line 56: @hugomrdias do you understand why `wantType` is omitted here ? Following lines make make me believe it should be provided, but I'm not sure what the value should be in this instance

https://github.com/ipfs/js-ipfs-bitswap/blob/681b015327a701834b377cce32edbb839cd709b5/src/types/wantlist/index.js#L28-L32

src/types/wantlist/index.js
```diff
@@ -72,9 +97,16 @@ class Wantlist {
   }
 
   sortedEntries () {
+    // TODO: Figure out if this is an actual bug.
+    // @ts-expect-error - Property 'key' does not exist on type 'WantListEntry'
```
Gozala, line 101: @achingbrain do you have an idea what this meant to be, I'm guessing key used to be what cid is today, but I'm not sure. Nor I have a clue how this supposed to be sorted.

src/types/message/index.js
achingbrain: Looks like a bug

src/types/wantlist/index.js
achingbrain: Looks like a bug.  I think this is just supposed to be a list of wantlist entries sorted by string CID, rather than insertion order which would be the default - the `.key` property access is probably due to misreading the docs on the output of `Map.entries()`?

I'm not sure what the value of it being sorted is, other than the output being stable - it's probably better as a set TBH.

Something we should create an issue for and maybe address later.

src/want-manager/index.js
achingbrain: I think the WantType ends up defaulting to `Block` if it's omitted.  The `Have`/`Block` distinction is a later optimisation, I think maybe it just got missed off here when it was implemented.

achingbrain (changes requested):
LGTM - a few bugs and inconsistencies are well pointed out but we should get this in and address those as follow ups.

Just needs the github URLs removing from deps and it's good to go.

src/network.js
Gozala: Created an issue in libp2p https://github.com/libp2p/js-libp2p/issues/899

Gozala:
Created an issue for all the discovered inconsistencies https://github.com/ipfs/js-ipfs-bitswap/issues/303

PR #286, opened 2021-01-19
fix: update provider multiaddrs before dial

achingbrain:
The output of `libp2p.contentRouting.findProviders()` is `AsyncIterable<{ id: PeerId, multiaddrs: Multiaddr[] }>`
and not `AsyncIterable<PeerId>` as the code seems to assume.

In order to ensure we can successfully dial a content provider we've not
dialled before, add their multiaddrs to the peer store before dialling them.

@vasco-santos is this the right place to do this?  Seems like it should happen in libp2p?

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -142,7 +142,9 @@ class Network {
     const connectAttempts = []
     for await (const provider of this.findProviders(cid, CONSTANTS.maxProvidersPerRequest, options)) {
       this._log('connecting to providers', provider.id.toB58String())
-      connectAttempts.push(this.connectTo(provider, options))
+
+      this._libp2p.peerStore.addressBook.set(provider.id, provider.multiaddrs)
+      connectAttempts.push(this.connectTo(provider.id, options))
     }
     await Promise.all(connectAttempts)
   }
--- a/src/network.js
+++ b/src/network.js
@@ -141,9 +141,7 @@ class Network {
   async findAndConnect (cid, options) {
     const connectAttempts = []
     for await (const provider of this.findProviders(cid, CONSTANTS.maxProvidersPerRequest, options)) {
-      this._log('connecting to providers', provider.id.toB58String())
-
-      this._libp2p.peerStore.addressBook.set(provider.id, provider.multiaddrs)
+      this._log(`connecting to provider ${provider.id}`)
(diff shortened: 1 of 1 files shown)


src/network.js
```diff
@@ -142,7 +142,9 @@ class Network {
     const connectAttempts = []
     for await (const provider of this.findProviders(cid, CONSTANTS.maxProvidersPerRequest, options)) {
       this._log('connecting to providers', provider.id.toB58String())
-      connectAttempts.push(this.connectTo(provider, options))
+
+      this._libp2p.peerStore.addressBook.set(provider.id, provider.multiaddrs)
+      connectAttempts.push(this.connectTo(provider.id, options))
```
vasco-santos, line 147: Libp2p should be responsible for this. In the DHT we add discovered multiaddrs to the addressBook when we do `findPeer`, but apparently not when we do `findProviders`. In the delegated side of things, we don't add them at all.

So, I think we should follow exactly the same approach as with the `peer-discovery` modules, which libp2p [adds to the AddressBook](https://github.com/libp2p/js-libp2p/blob/v0.30.1/src/index.js#L606). In this case, we should add the multiaddrs to the peerStore on [findProviders](https://github.com/libp2p/js-libp2p/blob/v0.30.1/src/content-routing.js#L64), [findPeer](https://github.com/libp2p/js-libp2p/blob/v0.30.1/src/peer-routing.js#L94) and [getClosestPeers](https://github.com/libp2p/js-libp2p/blob/v0.30.1/src/peer-routing.js#L123). We can later remove some redundant addressBooks adds on the DHT, but we need to be careful with this as DHT queries might discover important peers to store that might not be returned in these APIs.

What do you think?

A side note, we should use `add` here as we might already know other multiaddrs of the peer that we can use in the following connect. Only `identify` should use `set` at this moment, as this is the source of truth

achingbrain: I've added the storing of multiaddrs to findProviders here: https://github.com/libp2p/js-libp2p/pull/865 also made it a lot faster by returning results as they come in instead of waiting for one of the lookups to finish first.

I'll expand it to perform the same operations for findPeer and getClosestPeers tomorrow..

hugomrdias (approved):

PR #319, opened 2021-04-14
chore: replace protons with protobufjs

achingbrain:
Replaces protons with the official protobufjs module.


test/types/message.spec.js
```diff
@@ -25,19 +25,14 @@ describe('BitswapMessage', () => {
...
 
-      expect(Message.decode(serialized).wantlist.entries[0]).to.be.eql({
-        block: cid.bytes,
-        priority: 1,
-        cancel: false,
-        sendDontHave: false,
-        wantType: Message.Wantlist.WantType.Block
-      })
+      const deserialized = await BitswapMessage.deserialize(serialized)
+      expect(deserialized.wantlist.get(cid.toString())).to.have.nested.property('entry.wantType', Message.Wantlist.WantType.Block)
```
achingbrain, line 35: Protons did some non-standard things here - sendDontHave and wantType should come back as undefined but instead came back as `false` and `0` respectively in line with the comment in the .proto definition.

Protobufjs follows the spec and does not do this so we can't test against the output of the protobuf deserializer but neither should we, really as it's not the unit under test.

vasco-santos: protons was a mess, happy that we are getting rid of it

tsconfig.json
```diff
@@ -5,5 +5,8 @@
   },
   "include": [
     "src"
+  ],
+  "exclude": [
+    "src/types/message/message.js"
```
vasco-santos, line 10:
```suggestion
    "src/types/message/message.js" // generated file
```

vasco-santos (approved):
LGTM!
Pending CI failure on the bundlesize

achingbrain:
Three bytes!

image

PR #321, opened 2021-04-16
chore: fix types

achingbrain:
- Remove bignumber.js in favour of BigInt
- Use ipld-block in favour of the old ipfs-core-types
- Use blockstore from ipfs-repo instead of the old ipfs-core-types


package.json
```diff
@@ -97,10 +98,9 @@
   "dependencies": {
     "abort-controller": "^3.0.0",
     "any-signal": "^2.1.2",
-    "bignumber.js": "^9.0.0",
     "cids": "^1.1.6",
     "debug": "^4.2.0",
-    "ipfs-core-types": "^0.3.1",
+    "ipfs-core-types": "next",
```
vasco-santos, line 103: we need to get rid of these circular dependencies

src/types/wantlist/entry.js
```diff
@@ -41,15 +34,12 @@ class WantListEntry {
   }
 
   /**
-   * @param {API} other
-   * @returns {boolean}
+   * @param {any} other
```
vasco-santos, line 37:
```suggestion
   * @param {unknown} other
```

We are using unknown for such scenarios in libp2p. Perhaps better?

src/utils/index.js
```diff
@@ -6,7 +6,7 @@ const uint8ArrayEquals = require('uint8arrays/equals')
 /**
  * Creates a logger for the given subsystem
  *
- * @param {import('peer-id')} [id]
+ * @param {any} [id]
```
vasco-santos, line 9: can this be other than PeerID?

vasco-santos (approved):
LGTM
Couple of nits in review

src/utils/index.js
achingbrain: Oops, left that in by accident - I was toying with the idea of using CIDs instead of PeerIds

src/types/wantlist/entry.js
achingbrain: I don't think you can access properties of a variable of type `unknown`?

package.json
achingbrain: I don't think this is even required now, since we can pull `Blockstore` from `ipfs-repo` it's only in the tests, and even then it was only being used to pull types for the repo, which are in `ipfs-repo` now so it can go.

src/types/wantlist/entry.js
vasco-santos: We do it here https://github.com/libp2p/js-libp2p/blob/master/src/record/peer-record/index.js#L68
does not seem problematic

PR #323, opened 2021-04-20
fix: specify pbjs root

achingbrain:
If we do not specify a root for pbjs, the default is used which is
shared globally, so we can't have a protobuf message called `Message`
that exists for both `ipfs-bitswap` and `libp2p-kad-dht`, for example.

Specify a root to add a scope to the declaration of protobuf messages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "scripts": {
     "prepare": "npm run build",
     "build": "npm run build:proto && npm run build:proto-types && npm run build:bundle",
-    "build:proto": "pbjs -t static-module -w commonjs --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/types/message/message.js src/types/message/message.proto",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
   "scripts": {
     "prepare": "npm run build",
     "build": "npm run build:proto && npm run build:proto-types && npm run build:bundle",
-    "build:proto": "pbjs -t static-module -w commonjs --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/types/message/message.js src/types/message/message.proto",
+    "build:proto": "pbjs -t static-module -w commonjs -r bitswap --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/types/message/message.js src/types/message/message.proto",
```
achingbrain, line 36:
```suggestion
    "build:proto": "pbjs -t static-module -w commonjs -r ipfs-bitswap --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/types/message/message.js src/types/message/message.proto",
```

vasco-santos (approved):
This looks good. CI fails seem unrelated?

Should we add root names to every message to be safe?

achingbrain:
> Should we add root names to every message to be safe?

Yes, every use of pbjs should have a root added as this will bite us again..

PR #330, opened 2021-04-30
chore: update libp2p@0.31

vasco-santos:
Needs:
- [x] https://github.com/libp2p/js-libp2p/pull/935

### DIFF
--- a/test/decision-engine/decision-engine.js
+++ b/test/decision-engine/decision-engine.js
@@ -648,6 +648,7 @@ describe('Engine', () => {
 
     const onlyCases = []
     for (const testCase of testCases) {
+      // eslint-disable-next-line
       if (testCase.only) {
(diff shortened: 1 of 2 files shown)


test/decision-engine/decision-engine.js
```diff
@@ -648,6 +648,7 @@ describe('Engine', () => {
 
     const onlyCases = []
     for (const testCase of testCases) {
+      // eslint-disable-next-line
```
vasco-santos, line 651: This skips error link with ` testCase.only not permitted [Error/no-only-tests/no-only-tests]`

If I try to specify the actual error in the comment I get linting to shout that `Definition for rule 'no-only-tests' was not found. [Error/no-only-tests]` 🤷🏼

achingbrain (approved):

PR #336, opened 2021-05-12
fix: fixes unhandled promise rejection

evgkul:
Fixes https://github.com/libp2p/js-libp2p/issues/886


achingbrain:
Great catch, thanks for submitting this.  There were some build problems, I fixed them in #337 and added a test to prevent regressions but can't push to your branch to had to open a separate PR.

Superseded by #337

PR #342, opened 2021-05-26
Move process to dev dependencies

bgins:
Hello :wave:. We recently noticed that some modern bundlers moving away from shimming node built-ins (https://github.com/vitejs/vite/issues/2723#issuecomment-808690781), and that `process` is included as a dependency here. 

This PR would move `process` from `dependencies` to `devDependencies`. It appears to only be used in tests and benchmarks, and everything builds and tests pass afterwards.

I think this change would be alright, but certainly feel free to decline or request changes if I have overlooked something! :pray:


lidel:
@achingbrain sounds like you are the most familiar with this area of  js-ipfs?

jorishermans:
What is the process of this issue?

BigLep:
@Gozala : We're hoping during your tribute week you can review and merge.

achingbrain (approved):
Good catch, thanks!

